PR #1, opened 2019-06-17
release 136

robkooper:
This is what I received as part of the 136 revision. Looks like a lot of data was added and I can clean this up if needed.


robkooper:
The other option is I make 2 pull requests, one is just code+docs and the other is data.

billsacks:
Thanks a lot for this @robkooper !

A few questions based on a very quick initial review:

1. Should ArchiveVersions/historicalTrunk_0411.zip actually be kept?

2. My sense is that some of the added files here actually are just rearranged locations of files that were present before, so I expected some deletion of files. Examples are the various niwot.* files.

3. @davidjpmoore @jmzobitz - can one or both of you take a quick look to say (a) if this looks right to you, and (b) if one or both of you should be given credit as author of this commit (to the extent that it matters to you)

jmzobitz:
This looks correct to me.  Thanks for doing this!

billsacks:
@robkooper just pinging you again on my comments above: just want to make sure this doesn't fall through the cracks.

robkooper:
force pushed new updates, split into multiple commits
- movement of data around
- docs folder
- processFiles folder
- code changes

Hopefully this will make it easier to review so you can just look at a single commit to see the changes.

billsacks:
@robkooper thanks a lot. This is looking a lot better now. The one remaining issue I see is that the processingFiles folder (added in a6cba211531c7cf7d5c534759eab095a9860afaf) seems to duplicate a bunch of files. It looks like the relevant files should be deleted from their original location.

robkooper:
forced push a new set of commits that should have the files moved. Thanks for the quick checks @billsacks

billsacks:
Thanks a lot - looks good now!

jmzobitz:
Thanks all!

Issue #4, opened 2019-08-28
immediate evaporation values in smaller timesteps


istfer:
I am trying to calibrate SIPNET using half hourly LE fluxes. But I'm having issues with the immediate evaporation implementation with this small timestep.

Even with a reasonable 10% interception, immediate evaporation becomes too large due to this calculation [immedEvap = (rain) * params.immedEvapFrac](https://github.com/PecanProject/sipnet/blob/master/sipnet.c#L1370) 

Then the half hourly model output becomes too "noisy" with large numbers which is basically a scaled version of precipitation (see the figs below, sorry y-axes are different)
![normal](https://user-images.githubusercontent.com/13728388/63837443-27140400-c984-11e9-8935-f8df31751127.png)
![weird](https://user-images.githubusercontent.com/13728388/63837473-372be380-c984-11e9-8590-14eb61511e63.png)

This results in large LE values, especially in winter (where there are even no leaves to intercept rain if you're simulating a deciduous forest).

This is a problem for me because it affects calibration results. At some sites flux data is noisy and calibration favors the `immedEvapFrac` values that make the model outputs noisy too. But apart from my calibration problem, this is relevant for anyone who wants to use LE fluxes at 30 min steps.

I was discussing with @mdietze how to solve this. Here is a proposed solution for which I would like to have other SIPNET users'/developers' opinion:

1) Over larger timesteps this is less of an issue, so I guess the original implementation can stay. I was thinking of adding a flag, e.g. LEAF_WATER, which will have the functionality to turn on a more complex interception evaporation implementation. 

2) When LEAF_WATER is on, a maximum pool size will be calculated as a function of LAI and a new parameter which might be thought to be the thickness (depth) of the leaf rim, e.g. 
```
maxLeafPool = lai * leafPoolDepth
```

3) Then the (fixed) intercepted precipitation water will be calculated like `immedEvap = (rain) * params.immedEvapFrac` again, but if this term is greater than the pool size `maxLeafPool`, excess water will fall to the ground
```
    if(immedEvap > maxLeafPool)
      immedEvap = maxLeafPool
 // immedEvap is subtracted from rain later in the code
```

[...]

PR #5, opened 2019-09-05
temporary leaf evaporative pool

istfer:
First pass at calculation of a temporary leaf evaporative pool from which instantaneous evaporation can occur. 

Addressing [this issue](https://github.com/PecanProject/sipnet/issues/4).

### DIFF
--- a/sipnet.c
+++ b/sipnet.c
@@ -296,6 +296,7 @@ typedef struct Parameters {
   double m_ballBerry ; // slope for the Ball Berry relationship
   double leafCSpWt; // g C * m^-2 leaf area
   double cFracLeaf; // g leaf C * g^-1 leaf
+  double leafPoolDepth; // leaf (evaporative) pool rim thickness in mm
 
   double woodTurnoverRate; /* average turnover rate of woody plant C, in fraction per day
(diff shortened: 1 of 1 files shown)


sipnet.c
```diff
@@ -48,7 +48,7 @@
 // Lloyd-Taylor model is R = R0 * e^(E0 * (1/(-T0) - 1/(T - T0))) (see Lloyd and Taylor 1994, "On the temperature dependence of soil respiration")
 // where R0 is rate at 0 degrees C, T is soil temp., E0 and T0 are parameters
 
-#define SEASONAL_R_SOIL 0 && !LLOYD_TAYLOR
+#define SEASONAL_R_SOIL 1 && !LLOYD_TAYLOR
```
mdietze: Not sure why the addition of the leaf pool requires changing the definition of this variable

istfer: It doesn't hehe

I was testing with this flag myself, realized that the compiler gives errors due to missing parenthesis, thought I'd just add that fix as well. I can turn it back off

kzarada:
@mdietze I think this is ready to be merged? That way I can test out the LE fix

PR #11, opened 2023-04-04
Add support for Windows-style carriage return line feed in input files

olli4:
**Problem:** I got errors trying to run SIPNET by `make` and `./sipnet` in Windows Subsystem for Linux (WSL):

```
ERROR in readNamelistInputs: Read unexpected input item:

Please fix sipnet.in and re-run
```

**Cause:** The errors were because there was no support in SIPNET for Windows-style line changes: carriage return line feed, `\r\n`, in input files. Only `\n` line changes were supported. I'm not sure if the `\r` characters were introduced by Git ran in Windows or by SIPNET opening files in text mode in WSL.

**Fix implemented in PR:** I modified all the C sources to support `\r\n` line changes. The fixes were implemented such a way that they should not affect functionality when `\r` are not present in input files.

**Testing:** To test the modifications, I built SIPNET using `make` and ran `./sipnet` in WSL. The results in `Sites/Niwot/niwot.out` were identical to those from unmodified SIPNET ran on Linux (but somewhat different to those in the repo, for some reason; I updated the output file in the second commit). I also tested the transpose utility by `transpose Sites/Niwot/niwot.out` and it appeared to work without problems and the output looks correct. This still leaves some of my modifications untested and I don't know how to test those.


mdietze:
Could you confirm that the modified code produces the same outputs on a Linux machine as it did prior to your changes?

olli4:
> Could you confirm that the modified code produces the same outputs on a Linux machine as it did prior to your changes?

Yes, I just tried and `Sites/Niwot/niwot.out` from the unmodified and modified code look the same when running `./sipnet` in a Ubuntu Linux Docker container.

mdietze (approved):

PR #14, opened 2024-07-25
First draft of nitrogen and methane equations

dlebauer:
This is a very rough draft of equations required to simulate N2O and CH4.

I am not sure that latex + markdown was the best place to start, maybe pencil and paper or whiteboard would be better. But ... here we are! Definitely open to suggestions.

I wanted to follow the original formulation, but it uses T to mean both Temp and Transpiration; D for Day and Dependence function. This draft almost certainly contains inconsistent notation and many equations are placeholders. 

Here is a rendered version: https://rpubs.com/dlebauer/sipnet_equations

Since @Alomir will lead implementation, I think the next step will be for him and me to discuss what would be most useful.


dlebauer (commented):
Currently this is divided into "Current SIPNET" and "Proposed Updates". These should be combined into "Model structure" and each subsection "State variables", "Initial Values", etc should be separated into "Current" and "Proposed". This will require integrating original table into parameter lists, rewriting equations from Appendix A.

Perhaps we don't actually need to rewrite existing equations unless we want to update the notation.

Add list of drivers
- met
- irrig
- fertilization -
- tillage ... etc

Add equations required to represent agronomic events in the code

Biogeochemical Dynamics
- N Cycle
- Methanogenesis, 
- Both require estimation of % soil in anaerobic state 

Boundary conditions
- soil (water holding capacity ... anything else?)

State variables to add
- soil T
- litter C
- litter N
- Plant Wood and Leaf N: N_W, N_L (Analogous to C_W wood carbon)

Outputs:
- SOC
- N2O
- CH4

dlebauer:
Closing: I have moved all of this content to https://github.com/ccmmf/modeling_framework/tree/main/sipnet-model-description

Will move it back to this repository when it is ready

PR #15, opened 2024-08-08
Convert sipnet_inputs_outputs.doc to md

dlebauer:
Hopefully for easier reference. 

Exported sipnet_inputs_outputs.doc from google docs to md, then cleaned up formatting and added TOC


Alomir (commented):
I think a lot of the latex-style formatting is not working, due to misplaced $s? (That is, most everything in the rightmost column)

Alomir:
At least, I can't get it to look right in github (or vscode) - but I do notice that the equivalent formatting in [this PR](https://github.com/PecanProject/sipnet/pull/14) worked fine when exported to PDF.

dlebauer:
@Alomir it looks like a bug in GitHub's math processing. Since 1) it was only being used for degrees, times, and superscripts and 2) as discussed, this will eventually move to Doxygen inline docs, I just removed the latex formatting.

FWIW, Github markddown does support math https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions

But it has some bugs, apparently including that it doesn't like using `$^` for superscripts which on reflection is probably a misuse of the feature.

docs/sipnet_inputs_outputs.md
```diff
@@ -0,0 +1,232 @@
...
+| 58 | efficiency | conversion efficiency of ingested carbon |   |   |
+| 59 | maxIngestionRate | maximum ingestion rate of the microbe | hr-1 |   |
+| 60 | halfSatIngestion | half saturation ingestion rate of microbe | mg C g-1 soil |   |
+| 61 | totNitrogen | Percentage nitrogen in soil |   |   |
+| 62 | microbeNC | microbe N:C ratio | mg N / mg C |   |
+| 63 | microbeInit |   | mg C / g soil microbe | initial carbon amount |
+| 64 | fineRootFrac | fraction of wood carbon allocated to fine roots |   |   |
+| 65 | coarseRootFrac | fraction of wood carbon that is coarse roots |   |   |
+| 66 | fineRootAllocation | fraction of NPP allocated to fine roots |   |   |
+| 67 | woodAllocation | fraction of NPP allocated to the roots |   |   |
```
dlebauer, line 126:
```suggestion
| 67 | woodAllocation | fraction of NPP allocated to wood |   |   |
```

PR #17, opened 2024-08-16
16 Ensure CI execution of Niwot site

Alomir:
Adds an existence check for the test output

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,9 @@ jobs:
 
     # compile sipnet
     - name: compile sipnet
-      run: make
+      run: |
+        make
+        if (hashFiles('**/niwot.out') && )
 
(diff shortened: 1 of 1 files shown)


robkooper (changes requested):
The action seemed to fail on windows, which seems to be most of the work adding this module.

robkooper:
problem with the `rm -f` vs `rm` is that if the files does not exist it will fail with just the `rm` command. Seeing that it runs, I can only assume the file will exist.

robkooper (approved):

Alomir:
@robkooper Indeed, I'm not very excited by just `rm` - but at least in this case, all we want if the file to not exist when the test is run, so if it already doesn't, that's fine.

I'm hoping this is a stop-gap measure before we put a better regression test system in place.

PR #24, opened 2025-01-15
Move compile time flags

Alomir:
This PR moves two compile switches from `sipnet.c` to the new `modelStructures.h` file. No tests have been added for that functionality, but this should be done at some later date.

Also, it adds a script that can be run to update version of `modelStructures.h` in the unit test directories whenever changes are made to the "production" version. This script was run and the updated test version is in this PR.

### DIFF
--- a/tests/update_model_structures.sh
+++ b/tests/update_model_structures.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+# Run this script from the root SIPNET directory with the command:
+# > tests/update_model_structures.sh
+
+# Name of the file containing #define compiler switches
+STRUCTURES_FILE="modelStructures.h"
+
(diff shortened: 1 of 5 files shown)


tests/update_model_structures.sh
```diff
@@ -0,0 +1,70 @@
...
+        name = $2;
+        if (name in defines) {
+            printf "#define %s %s\n", name, defines[name];
+            next;
+        }
+    }
+    # Print the line as-is if no overwrite is needed
+    print $0;
+}
+EOF
```
infotroph, line 49: This is the most cleanly formatted and readable awk script I have ever seen. 💯

infotroph (approved):
This looks great, thanks! Extra points for managing to make a 3-language PR look easy.

PR #25, opened 2025-01-31
ORG78 Finish event wiring in SIPNET

Alomir:
This PR augments event handling in SIPNET by fetching events at the relevant place, and has a stubbed method for implementing the actual processing of events for later work.

Note that this is only for `runModelOutput()`, not `runModelNoOut()`; we can update the latter as well if desired later.

### DIFF
--- a/sipnet.c
+++ b/sipnet.c
@@ -489,8 +489,12 @@ static ClimateNode *climate; // current climate
 static Fluxes fluxes;
 static double *outputPtrs[MAX_DATA_TYPES]; // pointers to different possible outputs
 
+// MJL REMOVE THIS BEFORE CHECKIN
+#define EVENT_HANDLER 1
+
 #if EVENT_HANDLER
 static EventNode **events;
+static EventNode *locEvent; // current location event list
 #endif
 
 /* Read climate file into linked lists,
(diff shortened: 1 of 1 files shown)


sipnet.c
```diff
@@ -2275,54 +2276,103 @@ void updateTrackers(double oldSoilWater) {
...
-  		envi.snow += (fluxes.snowFall - fluxes.snowMelt - fluxes.sublimation) * climate->length;
-
-	#endif // MODEL_WATER
+  #endif // MODEL_WATER
 
   ensureNonNegativeStocks();
 
+#if EVENT_HANDLER
+  // Process events for this location/year/day, AFTER updates are made to fluxes and state
+  // variables above. Events are (currently, Jan 25) handled as instantaneous deltas to
```
dlebauer, line 2372:
```suggestion
  // variables above. Events are (currently, Jan 25, 2025) handled as instantaneous deltas to
```

dlebauer (approved):
Changes look great, looking forward to seeing how these events flesh out!

sipnet.c
```diff
@@ -2275,54 +2276,103 @@ void updateTrackers(double oldSoilWater) {
...
+        // TBD
+          printf("Tillage events not yet implemented\n");
+          break;
+      case FERTILIZATION:
+        // TBD
+          printf("Fertilization events not yet implemented\n");
+          break;
+      default:
+        printf("Unknown event type (%d) in processEvents()\n", locEvent->type);
+        exit(1);
```
robkooper, line 2321: do we really want to exit and not just print error message, and continue?

sipnet.c
```diff
@@ -2275,54 +2276,103 @@ void updateTrackers(double oldSoilWater) {
...
-
-  		// if COMPLEX_WATER = 0 or SNOW = 0, some or all of these fluxes will always be 0
-  		envi.snow += (fluxes.snowFall - fluxes.snowMelt - fluxes.sublimation) * climate->length;
-
-	#endif // MODEL_WATER
+  #endif // MODEL_WATER
 
   ensureNonNegativeStocks();
 
+#if EVENT_HANDLER
```
robkooper, line 2370: it seems that there are 2 places where you have #IF for the processEvents function (here and hte function itself), would it make sense to remove it from processEvents?

sipnet.c
```diff
@@ -2485,7 +2534,9 @@ void setupModel(SpatialParams *spatialParams, int loc) {
 
 // Setup events at given location
 void setupEvents(int currLoc) {
-  // Implementation TBD
+#if EVENT_HANDLER
+	locEvent = events[currLoc];
```
robkooper, line 2538: indenting looks off here

sipnet.c
```diff
@@ -2554,6 +2605,12 @@ void runModelNoOut(double **outArray, int numDataTypes, int dataTypeIndices[], S
 
   setupModel(spatialParams, loc);
 
+#if EVENT_HANDLER
+	// Implementation TBD
```
robkooper, line 2609: indenting looks off here

robkooper (changes requested):

sipnet.c
dlebauer: Formatting looks inconsistent throughout - what about applying a formatter in a separate PR?

sipnet.c
Alomir: That's a great question, I was going with what _seems_ to be the standard for SIPNET - that is, exit on errors that look like mistakes in inputs. Any thoughts on this @dlebauer @infotroph ?

sipnet.c
Alomir: Another great question - how careful do we want to be about trimming resulting code from all things EVENT_HANDLER when it is off? 

For the one in processEvents, that's the one that I lean towards keeping, as there's nothing stopping one from calling that function from somewhere else where there is no quard - and the contents of that function should not be run when EVENT_HANDLER is off. Feels safer.

Alomir: To be clear on that first part above, I'd leave the one in processEvents and remove the one around the call to processEvents - but that leaves a dummy call in the final code when off.

sipnet.c
Alomir: I cannot get CLion on Mac to listen to its own format settings, thus some wonky things sneak in. I def. vote for adding a formatter on checkin/push/merge or some such, but I'll fix this one here for now. Thanks for pointing that out!

sipnet.c
Alomir: Ditto

sipnet.c
Alomir: Decided via discussion: using different error codes, starting here

sipnet.c
Alomir: Decided via discussion: let's move the `#if` to surround the entire function - thus it would be a compile-time error if this function gets called outside of a `EVENT_HANDLER=1` scenarion

exitCodes.h
```diff
@@ -5,8 +5,14 @@
...
-  EXIT_CODE_UNKNOWN_EVENT = 2
+  EXIT_CODE_FAILURE = 1, // generic failure
+  // code 2 typically has a special meaning, see below
+  EXIT_CODE_UNKNOWN_EVENT = 3
 } exit_code_t;
 
+// Note for future: these codes typically have other meanings and should not be
+// used here (except for 0 and 1, for which our meaning is the special meaning)
+// 0-2, 126-165, and 255
+// See, e.g., https://tldp.org/LDP/abs/html/exitcodes.html
```
robkooper, line 16: other link that might be good is https://man7.org/linux/man-pages/man3/sysexits.h.3head.html

sipnet.c
```diff
@@ -2485,7 +2535,9 @@ void setupModel(SpatialParams *spatialParams, int loc) {
 
 // Setup events at given location
 void setupEvents(int currLoc) {
-  // Implementation TBD
+#if EVENT_HANDLER
```
robkooper, line 2538: probably should be consistent and have the EVENT_HANDLER around the function as well.

robkooper (changes requested):

sipnet.c
Alomir: Good catch!

exitCodes.h
Alomir: Or, in addition, looking at that more closely.

robkooper (approved):
one other thing to think about is to start a changelog, see pecan and https://keepachangelog.com/

PR #26, opened 2025-02-07
ORG79 Add irrigation event support

Alomir:
This PR:
* adds support for irrigation events (mostly in `sipnet.c` and `events.h/c`)
* simplifies handling of `modelStructures.h` for unit tests (see `Makefile`)
* adds a unit test for irrigation events

There is also a bit of clutter from small changes due to renaming and better `modelStructures.h` handling.

The 'spec' for this even type is from a slack thread discussion, snapshot here:
image


dlebauer:
RE: how do we want to test this? What should 'success' look like?


Perhaps, after reading an irrigation event:

1. Event is recognized and logged
   - irrig <0 --> error
  - printevent logs event
2. Event has intended impact
  - soilWater increases
  - if canopy irrigation: 
    - immedEvap increases

Alomir:
@dlebauer : I took the spirit of what you suggested and made a unit test out of it. (I believe the literal version would be a system test - which we need, but IMO systems tests should be their own discussion and effort.)

events.h
```diff
@@ -1,20 +1,21 @@
...
   FERTILIZATION,
   HARVEST,
   IRRIGATION,
   PLANTING,
   TILLAGE,
   UNKNOWN_EVENT
-};
+} event_type_t;
+
+typedef enum IrrigationLocation
```
dlebauer, line 13: in this context, the term 'location' is ambiguous - I was confused at first. The docs use the term 'type' but now I think 'method' would be better.

dlebauer (changes requested):
Hi Mike - I mean, the event_handler was cool and all, but this is  exciting! Not just that we can now irrigate, but because it is SIPNET's first steps toward supporting agriculture. 🎉 

I have a small request that I think will help clarity - changing 'irrigation location' to 'irrigation method'.

And I was trying to figure out where to put the specs that you pasted from slack - I'm happy to help, but was wondering if you had an opinion on whether it should be added to https://github.com/ccmmf/modeling_framework/pull/4, or whether it should be part of the in-code Doxygen documentation.

events.h
dlebauer: With apologies for being pedantic, I went ahead and implemented the nomenclature change (location-->method) in #32 and will also update the docs.

Alomir:
I put a fair amount of those docs into my changes here: [e660961](https://github.com/ccmmf/modeling_framework/pull/4/commits/e660961c53b1d0f4ce38001255169b7823cfe4ae) - which is in the PR you mentioned. I'm still thinking about how much we want in the code (definitely some).

dlebauer (approved):

PR #32, opened 2025-02-13
change 'irrigation location' to 'irrigation method'

dlebauer:


Alomir (approved):
I was not a fan of location, so this sounds great!

PR #36, opened 2025-03-02
feat: Automate SIPNET binary builds and manual release

### LINKED ISSUE #27: Automate SIPNET Binary Builds & Manual Release
## Description:

- #29 
- Upload binary builds for each (or at least the pre-ccmmf version and current one)
- Update PEcAn to access binaries from SIPNET repo instead of the ones currrenly hosted on google drive

Not essential, but could be automated with something like https://github.com/marketplace/actions/gh-release

## Motivations

1. Remove dependency of PEcAn CI on `make sipnet` success https://github.com/PecanProject/pecan/issues/3427.
2. Enable workflow users to download compile binaries, reducing overhead.

AritraDey-Dev:
This pull request fixes #27 

### CI Workflow Enhancements:

* [`.github/workflows/ci.yml`](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR66-R72): Added a step to upload SIPNET binaries after the build step.
* [`.github/workflows/ci.yml`](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR97-R125): Introduced a new job to create a release, which includes checking out the code, creating a release, and uploading the release asset.

### `Makefile` Updates:

* [`Makefile`](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L29-R29): Added a `release` target to the `all` target dependencies.
* [`Makefile`](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R54-R62): Defined the `release` target to package and compress the SIPNET binary and clean up the release directory.


Alomir (changes requested):
@AritraDey-Dev Thank you for the PR! 

Do I read this correctly, that it will create a release and upload binaries on every PR push? That would be way to often; at most, it should be on merging with the master branch.

dlebauer (changes requested):
Awesome, thanks for taking this on @AritraDey-Dev 

Two requests (with what I think the code would look like):

1. trigger on tags matching vX.Y.Z

```
on:
  push:
    tags:
      - 'v*.*.*'
```
2. create release as a draft 
```
steps:
  ...
  with:
    ...
    draft: true
```

AritraDey-Dev:
@dlebauer Thanks for the clarification.I will make those changes.

AritraDey-Dev:
> @AritraDey-Dev Thank you for the PR!
> 
> Do I read this correctly, that it will create a release and upload binaries on every PR push? That would be way to often; at most, it should be on merging with the master branch.

Yes, you're correct. The current setup would trigger a release on every PR push. We should modify it to only create a release and upload binaries upon merging into the master branch. Thank you for pointing that out!

dlebauer:
It looks like .github/workflows/ci.yml includes steps to both upload SIPNET binaries using actions/upload-artifact and to create a release and upload binaries using action-gh-release and upload-release-asset. Just want to confirm - are these redundant? If so, is that intentional?

dlebauer:
@AritraDey-Dev thanks for the clarification - I see a potetnial issue in that artifacts uploaded via `actions/upload-artifact` expire after 90 days https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period. 

One option would be to create a cron-job to make sure the artifacts are refreshed. Another option would be to use a build from the latest release - which is what I believe we decided in the PEcAn GHA.

.github/workflows/ci.yml
```diff
@@ -87,3 +96,33 @@ jobs:
       - name: Run Unit Tests
         run: make testrun
 
+  # Create Release
```
infotroph, line 99: i recommend moving this job to its own workflow file named `release.yml`, with just `on:push:tags' as the trigger event. Since triggers are per workflow, as written this will run on every PR and every push to master as well as on every tag event.

infotroph:
> used to make the binaries available within the workflow for other jobs (e.g., testing, debugging)

I'll defer to @Alomir here, but Sipnet compiles in less than a second so my sense is that compiling it separately for each job will be simpler and more reliable, and maybe faster too, than uploading and downloading the artifacts. Especially since the release job should only run on new tags while the testing jobs should run (with up-to-date binaries, not cached ones) on every push.

AritraDey-Dev:
> @AritraDey-Dev thanks for the clarification - to support the original use case motivation which was to create a binary that could be used in PEcAn workflows, [PecanProject/pecan#3427](https://github.com/PecanProject/pecan/issues/3427), I see a potential issue in that artifacts uploaded via `actions/upload-artifact` expire after 90 days https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period.
> 
> One option would be to create a cron-job to make sure the artifacts are refreshed. But in the PEcAn workflow we decided to use the latest stable release. [PecanProject/pecan#3427](https://github.com/PecanProject/pecan/issues/3427)
> 
> If there are other uses of the binary within the SIPNET CI workflows, then it certainly makes sense to keep both actions.

That makes sense. Since SIPNET CI workflows do rely on the binary, keeping both actions seems like the right approach. Setting up a cron job to refresh artifacts could help avoid expiration issues.

Makefile
```diff
@@ -26,7 +26,7 @@ DOXYFILE = docs/Doxyfile
 DOXYGEN_HTML_DIR = docs/html
 DOXYGEN_LATEX_DIR = docs/latex
 
-all: estimate sipnet transpose subsetData
+all: estimate sipnet transpose subsetData release
```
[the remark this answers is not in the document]
AritraDey-Dev: Thank you for your input! I agree that the release target should not be part of the all target. It makes sense to keep all focused on building all binaries in the source folder. I’ll update the Makefile accordingly to reflect this distinction

AritraDey-Dev:
@infotroph @dlebauer I've added a cron job and separated the release job into release.yml, triggering only on tag pushes to streamline the workflow. Let me know your thoughts!

.github/workflows/ci.yml
```diff
@@ -86,5 +86,4 @@ jobs:
 
       # run tests
       - name: Run Unit Tests
-        run: make testrun
-
+        run: make testrun
```
infotroph, line 89:
```suggestion
        run: make testrun

```

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'  # Trigger on version tags
+  schedule:
+    - cron: '0 0 * * *'  # Schedule cron
```
infotroph, line 8: I think you're mixing two incompatible suggestions here. @dlebauer was suggesting a cron job to keep build artifacts (which are temporary) from expiring, I was suggesting release and testing should be separate workflows that each compile their own binary and therefore do not need to create build artifacts in the first place.

I'll make some other suggestions below to show what I meant.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout code
```
infotroph, line 14: job name says this already
```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      # Download the SIPNET binary from the build job
+      - name: Download SIPNET Binary
+        uses: actions/download-artifact@v2
+        with:
+          name: sipnet-binaries
+          path: ./sipnet
```
infotroph, line 23:
```suggestion
      - name: Compile SIPNET binary
        run: make
```

Makefile
```diff
@@ -99,6 +99,7 @@ clean:
 	rm -f $(HISTUTIL_OFILES) $(COMMON_LIB) $(SIPNET_LIB)
 	rm -f estimate sensTest sipnet transpose subsetData bintotxt txttobin
 	rm -rf $(DOXYGEN_HTML_DIR) $(DOXYGEN_LATEX_DIR)
+	rm -f release/sipnet.tar.gz  # Clean up release package if it exists
```
infotroph, line 102:
```suggestion
```

Makefile
```diff
@@ -156,3 +157,13 @@ depend::
...
+# Target to run SIPNET with a specific input file
+run_sipnet: 
+	./sipnet -i tests/smoke/sipnet.in
+
+# Target to create a release package
+release: sipnet
+	mkdir -p release
+	cp sipnet release/
+	tar -czvf release/sipnet.tar.gz -C release sipnet
+	rm release/sipnet
```
infotroph, line 169: None of these outputs are used by the release steps above
```suggestion

```

infotroph (changes requested):

Makefile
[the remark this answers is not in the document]
Alomir: Thanks for that change - I was going to suggest the same

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/checkout@v2
+
+      # Build SIPNET binary
+      - name: Build SIPNET
+        run: make
+
+      # Create a GitHub Release (as a draft)
+      - name: Create Release
+        id: create_release
+        uses: softprops/action-gh-release@v1
```
Alomir, line 23: Are you using v1 intentionally? I believe v2 is the latest.

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          body: |
+            This release includes binary builds for historical SIPNET versions.
```
Alomir, line 28: Is this actually true? I think each release will only have one binary, and it's not historical - it's the latest and greatest :-)

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Build SIPNET
+        run: make
+
+      # Create a GitHub Release (as a draft)
+      - name: Create Release
+        id: create_release
+        uses: softprops/action-gh-release@v1
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
```
Alomir, line 26: I think `github.ref_name` would be better - we don't need the `refs/heads/` (or similar) prefix

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+          body: |
+            This release includes binary builds for historical SIPNET versions.
+          draft: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # Upload the SIPNET binary to the release
+      - name: Upload Release Asset
+        uses: actions/upload-release-asset@v1
```
Alomir, line 35: This action is unmaintained - I'm pretty sure the softprops action can do this too

Alomir (changes requested):

.github/workflows/release.yml
AritraDey-Dev: yes v2 is latest.i will update this.
Screenshot from 2025-03-12 20-49-32

Alomir:
Question for all - @AritraDey-Dev @dlebauer @infotroph - is there a reason we are doing a draft release? (Or am I overthinking this, and the intent is to remove that tag as a last step?)

AritraDey-Dev:
> Question for all - @AritraDey-Dev @dlebauer @infotroph - is there a reason we are doing a draft release? (Or am I overthinking this, and the intent is to remove that tag as a last step?)

yeah i share doubt though as @dlebauer told [here](https://github.com/PecanProject/sipnet/pull/36#pullrequestreview-2662650017) .
 However, my thinking is that drafting the release allows us to review the content, verify that the binaries are correct, and update any release notes before making it public. It’s essentially a quality control step to ensure we can make any last-minute adjustments before the release goes live

infotroph (approved):
Thank you! This looks good to me now. I'm OK either way with the draft-vs-not question, so let's get @dlebauer signoff on that.

AritraDey-Dev:
@infotroph @Alomir Thanks for approving!

PR #33, opened 2025-02-18
add changelog & update contributing

dlebauer:
Include information about SIPNET development from #29 

Created #46 to contain remaining tasks from draft in #34

PR #37, opened 2025-03-03
Reorganize codebase

Alomir:
Reorganization time!

Added several README's throughout, though they are still a work in progress.

Open issues/questions:
1. What, if anything, can we delete in the `archive` directory?
2. Is anyone using `estimate`?
3. Is anyone using the "other modes" of `sipnet` (`senstest` and `montecarlo`)?
  a. If not, can we trim `sipnet` down?
4. Documentation is scarce, and doesn't seem likely to fit in cleanly - but that all should be handled as a separate task
  a. By which I mean - we need (a) docs on how to build and run sipnet, including compile switches, and (b) there's doxygen output too. What goes where? Can we rely on doxygen for everything? Do we want to do that even if we can? Etc etc

Notes:
1. we are intentionally ignoring the linter report; a big part of doing this reorg now is to enable auto-format/linting for the code we actually care about.
2. @dlebauer 's version of [CONTRIBUTING.md](https://github.com/PecanProject/sipnet/blob/afff9d3a977adbcca65a8332e1e13ae8ccb7d98b/CONTRIBUTING.md) ([PR#33](https://github.com/PecanProject/sipnet/pull/33)) has some of the info referred to by my TODOs at the top of that doc


dlebauer:
I think 1-3 is mostly for @jmzobitz and any other interested contributors to the original SIPNET. 

A key objective of this reorganization is to make the codebase easier to maintain, use, and document as we develop new functionality. The PEcAn / CCMMF team priority is maintaining and extending the simple forward case - and relying on PEcAn for analysis (including ensembles, fitting, sensitivity analysis, data assimilation). Key ? for @jmzobitz is whether you use or plan to use these tools and if so, are you interested in maintaining them as we develop new functionality? If not, would you prefer that we maintain the in the repository in an archive/ folder vs removing them - noting that they will always be available in versions <=1.3.0.

I also want to take the chance to complement the team on the implementation and application of these algorithms. It is nicely done, and the functionality of SIPNET as a combination of model + these analysis is a really cool feature. And bonus points for doing this all in C.

jmzobitz:
Sorry - just getting caught up on emails here.  Thanks for the tag.

I am 100% in the camp that let's launch SIPNET forward with the new tools that you are all developing.  Meaning, I see no need in maintaining the old `estimate` `montecarlo`, and `senstest`, and can adapt my work to the new builds you are making.

In terms of archiving code, I have a historical version saved locally of all the old open issues, so if there is ever any need to go back and to reanalyze scripts from previous publications, that can be done.

I am viewing this as an opportunity for me to put some fresh eyes on my coding skills.  :-)

CONTRIBUTING.md
```diff
@@ -1,6 +1,13 @@
 
 ## Style Guide
 
+TODO:
+1. add some 'code of conduct' lines here, or a separate file?
```
dlebauer, line 5: #33

dlebauer (changes requested):
Nothing like some spring cleaning!

Suggestions:
- Remove the archive with a fond farewell 🫠
- Merge this and #33 and then I can update the draft README

dlebauer (approved):
removed archive directory in https://github.com/PecanProject/sipnet/pull/37/commits/a95aca5501fab8c640443297b33cd69d4e60f2e7; ready to merge

PR #42, opened 2025-03-09
Migrate model structure and parameter docs from modeling framework repo to SIPNET repo

dlebauer:
### Moved docs from the modeling_framework repo while preserving git history:

- model-structure.md
- parameters.md
- .alternate-model-structure-ideas.md

Renamed `sipnet_inputs_outputs.md` to `.old-parameters.md`; we should be able to delete this after we finish parameters.md

## Key Changes

**It is not necessary to review all of the contents** Diff of changes since the last review can be viewed in https://github.com/ccmmf/modeling_framework/pull/6/files

- Updated planting and harvest specifications as discussed Friday 
- make index variables (i,j) clearly and consistently defined
- fixed eq 26 C = LAI *  SLW (instead of incorrect LAI/SLW)
- Updated formatting to distinguish existing vs proposed features:

---

(I sacrificed readability for the opportunity to use this font)

image
image
image


docs/parameters.md
```diff
@@ -0,0 +1,560 @@
...
+  - for perennials, some biomass may remain (col 5 + col 7 <= 1 and col 6 + col 8 <= 1; remainder is living).
+   - root biomass is only removed for root crops
+ 
+#### Example of events file:
+
+```
+1  2022  40  irrig  5   1        # 5cm canopy irrigation on day 40
+1  2022  40  fert   100 50 0 0   # fertilized with 10 g / m2 NO3-N and 5 g / m2 NH4-N on day 40
+1  2022  35  till   0.2 0.3      # tilled on day 35, soil organic matter pool decomposition rate increases by 20% and soil litter pool decomposition rate increases by 30% 
+1  2022  40  plant  10 10 1      # planting - planting occurs on day 40, after 10 days the plants have emerged with a size of 10 g C/m2 and 1 g N / m2
```
Alomir, line 502: We should update this plant event - I assume it would just be this?
```suggestion
1  2022  50  plant       # planting - emergence occurs on day 50
```

docs/parameters.md
```diff
@@ -0,0 +1,560 @@
...
+| litter decomposition modifier |  6  |  Y   | % increase in $K_{lit}$  |
+
+  - model representation:
+    - increase k for one month, amount proportional to depth
+    - transfer litter C and N to soil pool
+  - notes: could also alter bulk density and other soil properties
+
+#### Planting Events
+
+- model representation:
```
Alomir, line 475: Should we explicitly point out that a planting event has zero parameters (other than the type itself, of course)?

Alomir (approved):
I think this is great as is - we can fix things as we find them.

A couple of minor comments around planting events, but go ahead and merge once those are taken care of.

docs/parameters.md
dlebauer: yep thanks for catching that!

PR #45, opened 2025-03-10
SIP38 Fix OOM issues when reading bad data

Alomir:


infotroph (approved):
Hooray for immediate well-labeled failures!

PR #52, opened 2025-03-25
SIP51 Allow empty or nonexistent event files

Alomir:
This PR changes `readEventData` to allow the events file to be empty or nonexistent, and treat it as "no events specified", even when `EVENT_HANDLER` is on.

### DIFF
--- a/src/sipnet/events.c
+++ b/src/sipnet/events.c
@@ -160,8 +160,6 @@ EventNode **readEventData(char *eventFile, int numLocs) {
   char line[EVENT_LINE_SIZE];
   EventNode *curr, *next;
 
-  printf("Begin reading event data from file %s\n", eventFile);
-
   EventNode **events =
       (EventNode **)calloc(sizeof(EventNode *), numLocs * sizeof(EventNode *));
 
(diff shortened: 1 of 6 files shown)


src/sipnet/events.c
```diff
@@ -163,13 +164,21 @@ EventNode **readEventData(char *eventFile, int numLocs) {
...
   EventNode **events =
       (EventNode **)calloc(sizeof(EventNode *), numLocs * sizeof(EventNode *));
-  // status of the read
+
+  // Check for a non-empty file
+  if (access(eventFile, F_OK) != 0) {
+    // no file found, which is fine; we're done, a vector of NULL is what we
+    // want for events
+    return events;
+  }
```
infotroph, line 173: Would it make sense to move the "begin reading..." printf below this, so that when the file doesn't exist it doesn't report reading from it? (I think if sipnet.event is present but empty the reading message is fine, it just surprised me to see it when the file didn't exist)

infotroph (approved):
👍 👍 Thanks! I see one possible tweak on message output (See inline), but I won't need to re-review.

src/sipnet/events.c
Alomir: Good point! I think I'm going to also report when there is no file, with a "assuming no events" blurb.

Issue #41, opened 2025-03-06
Compile sipnet docs to gh-pages


dlebauer:
### Description:

- compile all docs (md + doxygen) into a single website
- create GitHub action to publish these to gh-pages when changes are merged into master branch
- optional - add make document-website (?)

In the short term, start with what we have (README, CONTRIBUTING, Doxygen, CODE_OF_CONDUCT, CHANGELOG from #33)

In the end, the documentation structure could be:

``` 
├── Home (README)
├── User Guide
│   ├── Running SIPNET
│   ├── Model Structure
│   └── Parameters and Configuration
├── Developer Guide
│   ├── Code structure
│   └── Contributing (Contributing.md)
│     └── additional details beyond scope of Contributing.md
├── API Reference (generated from Doxygen)
└── Changelog (if just a matter of adding another md file)
└── Code of Conduct
```

### Approach

As far as frameworks, it looks like MkDocs is simple and lightweight which is consistent with the SIPNET approach. I also considered Docusaurus since it is used for the website (https://github.com/pecanproject/pecanproject.github.io), but I don't think the additional overhead is warranted for a straightforward documentation site. Both will combine md and doxygen html. 

Not required, but consider whether [MkDoxy](https://mkdoxy.kubaandrysek.cz/#feature-list) will simplify Doxygen integration.

AritraDey-Dev:
@dlebauer #50  would it help?

PR #53, opened 2025-03-26
Build documentation using MkDocs (replaces #41)

shah0108:
All the things are added only one thing 1 thing I had confusion about was `CHANGELOG.md` to be added manually or not but that was not present already present so I created it, please take look at it if something needs to get modified please tell me, thanks!


dlebauer:
It appears that the Doxyfile (or a related path configuration) is causing Doxygen to scan more directories than intended. My sipnet/ directory is located alongside pecan/ and pecan.github.io/, and when I run make document, Doxygen processes a large number of graphs unnecessarily.

for example:

```
$ pwd
/path/to/sipnet

$ ls -d ../
sipnet/  pecan/  pecan.github.io/

$ make document
# generates >10k files along with output including:
...
Running dot for graph 3870/22969
Running dot for graph 3871/22969
Running dot for graph 3872/22969
... 
```

When I inspect sipnet/docs/html/files.html, it includes files from directories other than sipnet/ (including pecan/, pecan.github.io/ ...).

Please adjust the Doxyfile and/or related configuration so that only the sipnet/ directory is processed. This should improve build time and keep the generated documentation focused on SIPNET.

Makefile
```diff
@@ -124,7 +131,7 @@ $(SIPNET_TEST_DIRS_CLEAN):
 
 cleanall: clean testclean
 
-.PHONY: all clean histutil help document .doxygen.stamp exec cleanall \
+.PHONY: all clean histutil help document exec cleanall \
```
dlebauer, line 134: I was expecting that `make clean` would remove the entire `sites/` - should it? If so, can you please add?

shah0108:
> It appears that the Doxyfile (or a related path configuration) is causing Doxygen to scan more directories than intended. My sipnet/ directory is located alongside pecan/ and pecan.github.io/, and when I run make document, Doxygen processes a large number of graphs unnecessarily.
> 
> for example:
> 
> ```
> $ pwd
> /path/to/sipnet
> 
> $ ls -d ../
> sipnet/  pecan/  pecan.github.io/
> 
> $ make document
> # generates >10k files along with output including:
> ...
> Running dot for graph 3870/22969
> Running dot for graph 3871/22969
> Running dot for graph 3872/22969
> ... 
> ```
> 
> When I inspect sipnet/docs/html/files.html, it includes files from directories other than sipnet/ (including pecan/, pecan.github.io/ ...).
> 
> Please adjust the Doxyfile and/or related configuration so that only the sipnet/ directory is processed. This should improve build time and keep the generated documentation focused on SIPNET.

hello sir, i m not getting the pin point of the conversation but I tried to figure it out first of all the `INPUT` must be set to `.` since it contains `. . /` which might be the reason it is considering other files too
image

I think above changes should solve the issue but to be specific we can use `exclude : ../pecan ../pecan.github.io` and `exclude patterns : */pecan/* */pecan.github.io/* ` 
image

I'll make the above changes also can you help me if the changes I mentioned are the ones, like I did get what issue is but I've not much knowledge on how it would be solved but i guess with the above changes it should be done?

Makefile
shah0108: > I was expecting that `make clean` would remove the entire `sites/` - should it? If so, can you please add?

image
here in the clean i guess i need to also add `clean: rm -rf sites/` if I am not wrong?

AritraDey-Dev:
@shah0108 instead of excluding i think doing `INPUT                  = ../src`   will fix it.
As if you exclude some specific projects it may include some other projects `makefile` from the dir. Can you also try this way once...

Makefile
AritraDey-Dev: > > I was expecting that `make clean` would remove the entire `sites/` - should it? If so, can you please add?
> 
> image here in the clean i guess i need to also add `clean: rm -rf sites/` if I am not wrong?

`rm -rf site/` instead of ` rm -rf sites/`

dlebauer:
Sorry if it wasn't clear but excluding directories that are specific to my system doesn't seem like the right answer. I think @AritraDey-Dev is on the right track if the paths is relative to the doxyfile.

shah0108:
> @shah0108 instead of excluding i think doing `INPUT = ../src` will fix it. As if you exclude , it may include some other projects `makefile` from that dir. Can you also try this way once...

Yes makes sense I tried and is working as usual, would commit now the changes.

shah0108:
> Sorry if it wasn't clear but excluding directories that are specific to my system doesn't seem like the right answer. I think @AritraDey-Dev is on the right track if the path is relative to the doxyfile.

Yes sir it won't work but, I was only giving an example, that we can exclude all those files which also have `Makefile` but now since I got the correct way it is not required anymore, thanks!

AritraDey-Dev:
Hi @shah0108, I tested your code and deployed the documentation using GitHub Pages. While the documentation does generate and display the main page correctly, there are some issues with the API reference.

Currently, the "API Reference" section only displays the main page without the expected structured navigation (e.g., files, classes, functions, namespaces, etc.). Ideally, when viewing the documentation—either locally or on GitHub Pages

the api reference page should look some thing like this
Screenshot from 2025-04-06 21-41-22

shah0108:
Hello @AritraDey-Dev see I can explain but videos would be more preferable, let me explain first

Initially when I made the `docs site` for the first time I used a separate page `index.md` for showing it and then a link which could redirect it to this `SIPNET documentation` but due to `README.md` it was showing error and therefore then I made a separate `home/` directory for the `README.md` then everything was running fine but @dlebauer sir told me to remove the extra directory and directly link the API reference to `index.html` and therefore it is showing direct linkage to the SIPNET documentation! Did you get it what I meant?

mkdocs.yml
```diff
@@ -0,0 +1,17 @@
+site_name: SIPNET Documentation
+
+nav:
+  - Home: README.md
+  - User Guide:
+      - Running SIPNET: user-guide/running-sipnet.md
+  - Developer Guide:
+      - Code Structure: developer-guide/code-structure.md
+  - Contributing: CONTRIBUTING.md
+  - API Reference: html/index.html
```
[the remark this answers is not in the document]
AritraDey-Dev: Doxygen-generated API docs should be in `api` folder.

shah0108:
> Hi @shah0108, I tested your code and deployed the documentation using GitHub Pages. While the documentation does generate and display the main page correctly, there are some issues with the API reference.
> 
> Currently, the "API Reference" section only displays the main page without the expected structured navigation (e.g., files, structures, etc.). Ideally, when viewing the documentation—either locally or on GitHub Pages
> 
> the api reference page should look like this Screenshot from 2025-04-06 21-41-22

Hey @AritraDey-Dev sorry for the previous reply I was in a bit hurry with college exams and submissions and all and I completely misunderstood what you mean to say 
[Screencast from 2025-04-07 01-03-54.webm](https://github.com/user-attachments/assets/da3b1ccb-5540-4c98-ae3a-72b3efd76a10)
please refer the video once, it is showing the same thing as you said it must show in `api reference`, isnt it??

shah0108:
> Hi @shah0108, I tested your code and deployed the documentation using GitHub Pages. While the documentation does generate and display the main page correctly, there are some issues with the API reference.
> 
> Currently, the "API Reference" section only displays the main page without the expected structured navigation (e.g., files, structures, etc.). Ideally, when viewing the documentation—either locally or on GitHub Pages
> 
> the api reference page should look like this Screenshot from 2025-04-06 21-41-22

Hey @AritraDey-Dev sorry for the previous reply due to college exams and submissions I just didn't read your whole issue and just replied on what I thought you wanted to say, I just noticed your issue was completely different then what I thought sorry for that and yes it is only showing the main page, I don't know why is this happening?

mkdocs.yml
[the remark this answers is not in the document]
shah0108: > @dlebauer I was under the impression that placing the Doxygen-generated API docs in the docs/api directory is the expected approach for MkDocs to render them properly — especially under the "API Reference" section. Does that sound right to you?

Can you please explain a bit, basically what you mean to say is I need to create a new directory named `API Refrence` and under that directory, I should move the `html` directory and then in *mkdocs.yml* I should address them properly right??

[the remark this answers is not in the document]
AritraDey-Dev: Basically In the **API Reference** section, the expected content is not being displayed correctly.

I have investigated the issue and identified the root cause. I will suggest the necessary changes below to address and resolve this.

docs/Doxyfile
```diff
@@ -965,7 +967,7 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = ../
+INPUT                  = ../src
```
AritraDey-Dev, line 970:
```suggestion
INPUT                  = src
```

AritraDey-Dev: The issue was that the API Reference section appeared empty because the Doxygen INPUT path was incorrectly set to ../src, which did not exist relative to the project root; changing it to src resolved the issue by correctly pointing to the actual source files.

AritraDey-Dev (changes requested):

AritraDey-Dev:
Hi @dlebauer,

While running` make document` after performing a `make clean`, I encountered the following warning and message:
```
make document 
find: warning: you have specified the global option -mindepth after the argument -type, but global options are not positional, i.e., -mindepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
find: warning: you have specified the global option -maxdepth after the argument -type, but global options are not positional, i.e., -maxdepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
make: Nothing to be done for 'document'.
```
I found that this behavior is caused by the presence of the `.doxygen.stamp` and `.mkdocs.stamp` files in the project directory. These timestamp files appear to prevent the document target from rebuilding the documentation, even after a make clean.

Would it be appropriate to include the removal of these files in the clean target? For example:
`rm -f .doxygen.stamp .mkdocs.stamp
`

shah0108:
> @dlebauer @shah0108
> 
> While running` make document` after performing a `make clean`, I encountered the following warning and message:
> 
> ```
> make document 
> find: warning: you have specified the global option -mindepth after the argument -type, but global options are not positional, i.e., -mindepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
> find: warning: you have specified the global option -maxdepth after the argument -type, but global options are not positional, i.e., -maxdepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
> make: Nothing to be done for 'document'.
> ```
> 
> I found that this behavior is caused by the presence of the `.doxygen.stamp` and `.mkdocs.stamp` files in the project directory. These timestamp files appear to prevent the document target from rebuilding the documentation, even after a make clean.
> 
> Would it be appropriate to include the removal of these files in the clean target? For example: `rm -f .doxygen.stamp .mkdocs.stamp `

Hey @AritraDey-Dev , thanks for bringing this up!

I haven’t run into this issue on my end—after doing a `make clean`, running `make document` works just fine for me. But yeah, I get how those `.doxygen.stamp` and `.mkdocs.stamp` files could be causing problems in certain setups.

It makes sense to include their removal in the clean target to keep things consistent across environments. We could either add that to the current clean rule or maybe create a `distclean` target for a more thorough cleanup. What do you think @dlebauer @AritraDey-Dev ?

AritraDey-Dev:
@shah0108 I believe removing them with `rm -f .doxygen.stamp .mkdocs.stamp` in `make clean` shouldn’t cause any issues and would solve the problem mentioned—but let’s get @dlebauer’s sign-off on that.

Makefile
```diff
@@ -99,6 +106,7 @@ clean:
 	rm -f $(HISTUTIL_OFILES) $(COMMON_LIB) $(SIPNET_LIB)
 	rm -f estimate sensTest sipnet transpose subsetData bintotxt txttobin
 	rm -rf $(DOXYGEN_HTML_DIR) $(DOXYGEN_LATEX_DIR)
+	rm -rf site/
```
dlebauer, line 109:
```suggestion
	rm -rf site/
	rm -f .doxygen.stamp .mkdocs.stamp
```

dlebauer:
@shah0108 and @AritraDey-Dev 

- I've updated make clean to remove the stamp files. 
- Regarding this thread https://github.com/PecanProject/sipnet/pull/53#discussion_r2031681717, please check if https://github.com/PecanProject/sipnet/pull/53/commits/2996af07e517248723a627458db5d533dcefd9fd resolves the issue? Doxygen html is now written to api/index.html. And I added src/README.md to the Doxygen page so that it isn't blank. 

In the future we could better integrate Doxygen and mkdocs, and add the parameters and model specification to the mkdocs site, but as is, this PR is a significant improvement on what we had for docs so if this is ready, lets merge!

dlebauer (approved):

shah0108:
thanks for the merge @dlebauer sir!🎉

PR #55, opened 2025-03-27
ORG93 Handle planting and harvest events

Alomir:
https://github.com/ccmmf/organization/issues/93

In this PR:
* adds to sipnet.c to process planting and harvest events
* adds tests for such
* updates parameters.md for new planting spec
* adds explicit `coarseRootAllocation` parameter, and updates `ensureAllocation()` to better check allocation params

Note: creating `events.out` for reporting event handling data will be done in a follow-on task (likely the next one).


src/sipnet/sipnet.c
```diff
@@ -172,6 +172,9 @@
 
 #define TINY 0.000001  // to avoid those nasty divide-by-zero errors
 
+// For planting events
+#define LAI_AT_EMERGENCE 0.15
```
infotroph, line 176: Don't yet see where this is scaled  to peak LAI (which is what it's 15% of)

Alomir: But, perhaps my understanding of LAI is incorrect - I thought it was Leaf Area Index, the ratio of leaf area to total area (thus unitless, or perhaps more accurately m^2/m^2); this quantity times leaf specific weight (g/m^2) gives us our g/m^2 C quantity for the leaves.

infotroph: Your understanding of LAI is right, but it looks like we missed a word in the model equations. What we said in the monitoring framework was (my emphasis):

> The SIPNET model will be able to use the effective date directly by "planting" vegetation with leaf biomass set to **15% peak** and stem and root biomass set allometrically from leaf.

So I think we missed the "peak" part in the model equations... though it is true what the phenology product actually measures is 15% of peak _greenness_, not LAI directly.  @dlebauer am I missing some way peak greenness can be simplified down to a constant? I know all vegetation indexes saturate, but I think they all do it at LAI higher than 1.

Alomir: Oh, so it's not a straight `0.15` factor, but needs to be modified by the peak greenness?

infotroph: Recap of Slack discussion: We definitely got the spec wrong here. Updates coming; short version is that planting events will need to include the amounts of C added to leaf, stem, fine root, and coarse root pools. Converting from other units (e.g. LAI) to determine those carbon pools will be up to the user.

src/sipnet/sipnet.c
```diff
@@ -375,7 +375,10 @@ typedef struct Parameters {
   double woodAllocation;  // fraction of NPP allocated to the roots
   double fineRootExudation;  // fraction of GPP exuded to the soil
   double coarseRootExudation;  // fraction of NPP exuded to the soil
-  // 6 parameters
+  // Calculated param
+  double coarseRootAllocation;  // fraction of NPP allocated to the coarse roots
```
dlebauer, line 379: I think that `coarseRootFrac` is the parameter that needs to be added so that it can be initialized at emergence.

docs/model-structure.md
```diff
@@ -616,29 +616,9 @@ The choice of one month adjustment period is based on DayCent (Parton et al 2001
...
-
-$$
-C_{i,e} = C_{\text{plant,e}}\cdot \alpha_i \tag{27} \\
-$$
-
-$$\small i \in \{\text{leaf, wood, fine root, coarse root}\}$$
-
-
-Finally, nitrogen for each pool is calculated using equation \eqref{eq:cn_stoich}.
+Following carbon addition, nitrogen for each pool is computed using the corresponding C:N stoichiometric ratios followingn equation \eqref{eq:cn_stoich}.
```
dlebauer, line 621:
```suggestion
Following carbon addition, nitrogen for each pool is computed using the corresponding C:N stoichiometric ratios following equation \eqref{eq:cn_stoich}.
```

docs/parameters.md
```diff
@@ -473,10 +474,16 @@ Notes:
 
 #### Planting Events
 
+| parameter     | col | req? | description                                  |
+|---------------|:---:|:----:|----------------------------------------------|
+| leaf-C        |  5  |  Y   | C added to leaf ppol (g C / m2)              |
```
dlebauer, line 479:
```suggestion
| leaf-C        |  5  |  Y   | C added to leaf pool (g C / m2)              |
```

docs/parameters.md
```diff
@@ -125,6 +125,7 @@ Run-time parameters can change from one run to the next, or when the model is st
 |   | ${N_2O}_{\text{soil},0}$                         |                 | Nitrous oxide concentration in the soil   | g N m$^{-2}$                 |                                                   |
 |   | $C_{\text{fine root},0}$    |  fineRootFrac    | Initial fine root carbon                  | g C m$^{-2}$ | |
 |   | $C_{\text{coarse root},0}$  |   coarseRootFrac | Initial coarse root carbon                | g C m$^{-2}$                 |                                                   |
+         |                                                   |
```
mdietze, line 128: not clear what these are for

Alomir: I think these should be "fraction of `plantWoodInit` that is allocated to {fine|coarse} root carbon pool"

I'm happy to update these, but I'd like to forestall this becoming a general review of parameters.md, if that's ok

docs/parameters.md
```diff
@@ -111,20 +111,19 @@ Run-time parameters can change from one run to the next, or when the model is st
...
+|    | $N_{\text{soil},0}$        |                 | Initial soil organic nitrogen content                                    | g N m$^{-2}$                                         |                                                                                    |
+|    | ${CH_4}_{\text{soil},0}$   |                 | Initial methane concentration in the soil                                | g C m$^{-2}$                                         |                                                                                    |
+|    | ${N_2O}_{\text{soil},0}$   |                 | Nitrous oxide concentration in the soil                                  | g N m$^{-2}$                                         |                                                                                    |
+|    | $C_{\text{fine root},0}$   | fineRootFrac    | Fraction of `plantWoodInit` allocated to initial fine root carbon pool   | g C m$^{-2}$                                         |                                                                                    |
```
dlebauer, line 125:
```suggestion
|    | $C_{\text{fine root},0}$   | fineRootFrac    | Fraction of `plantWoodInit` allocated to initial fine root carbon pool   |                                      |                                                                                    |
```

docs/parameters.md
```diff
@@ -111,20 +111,19 @@ Run-time parameters can change from one run to the next, or when the model is st
...
+|    | $C_{\text{coarse root},0}$ | coarseRootFrac  | Fraction of `plantWoodInit` allocated to initial coarse root carbon pool | g C m$^{-2}$                                         |                                                                                    |
```
dlebauer, line 126:
```suggestion
|    | $C_{\text{coarse root},0}$ | coarseRootFrac  | Fraction of `plantWoodInit` allocated to initial coarse root carbon pool |                                  |                                                                                    |
```

dlebauer (approved):

dlebauer:
Looks good - I just updated [fine|coarse]RootFract definitions so that they are represented by $f$ (dimensionless fractions) instead of $C$ (pool with mass/area)

docs/parameters.md
dlebauer: Of course ... @mdietze, a separate general review of parameters.md and model-specification.md are welcome, and may help make sure we are all on the same page with the model! 🙏🏻

PR #56, opened 2025-03-31
Tweaks to hook setup

infotroph:
Trying to fix a couple points I stumbled on while setting up pre-commit hooks. All strictly "worked for me, maybe there's a better way" -- edits welcome.


tools/clang-format.hook
```diff
@@ -10,15 +10,15 @@ read -r -d '' help <<- EOF
...
 problems occur:
-$ brew install git-clang-format
-or
 $ sudo apt-get remove clang-format-*
 $ sudo apt-get install clang-format-19
 You should now have clang-format-19 installed. To ensure that git finds this binary,
 update the configuration as follows:
 $ git config --global clangFormat.binary clang-format-19
 $ git config --global clangFormat.style file
 $ git config --global alias.clang-format clang-format-19
```
infotroph, line 19: Would it make sense to add (the correct mac-specific variants of) these git config calls to setup.sh? I'm not clear how much these are universal vs. needing adjustment for each user.

Alomir: I thought those were valid on mac? Did you run into problems with them?

tools/clang-format.hook
```diff
@@ -10,15 +10,15 @@ read -r -d '' help <<- EOF
 This repository requires you to install the clang-format command. This hook has
 been tested with clang-format-19, so remove all earlier versions and reinstall if
 problems occur:
-$ brew install git-clang-format
-or
```
Alomir, line 14: Why remove these? (Or, maybe we should state 'this is for Mac'?)

tools/setup.sh
```diff
@@ -70,6 +70,7 @@ if [[ ! "$found_format" || ! "$found_tidy" ]]; then
     brew install llvm
     ln -s "$(brew --prefix llvm)/bin/clang-format" "/usr/local/bin/clang-format"
     ln -s "$(brew --prefix llvm)/bin/clang-tidy" "/usr/local/bin/clang-tidy"
+    ln -s "$(brew --prefix llvm)/bin/git-clang-format" "/usr/local/bin/git-clang-format"
```
Alomir, line 73: Good catch, thanks!

tools/clang-format.hook
Alomir: Ah, nevermind, I see!

tools/clang-format.hook
Alomir: I *think* the llvm install of git-clang-format must handle this, as I didn't need to futz with my config? Did you end up needing to do so?

infotroph: I just unset all three of them and best I can tell the hook is still working as expected, so apparently I did not need to.

Alomir (approved):
Looks good, definitely an improvement even if not in a "final state".

PR #62, opened 2025-04-15
Add soilWaterInit to enable the direct assimilation of volumetric soi…

Qianyuxuan:
…l moisture

### DIFF
--- a/src/sipnet/sipnet.c
+++ b/src/sipnet/sipnet.c
@@ -233,6 +233,7 @@ typedef struct Parameters {
   double soilInit;  // g C * m^-2 ground area
   double litterWFracInit;  // unitless: fraction of litterWHC
   double soilWFracInit;  // unitless: fraction of soilWHC
+  double soilWaterInit;  // cm soil water
   double snowInit;  // cm water equiv.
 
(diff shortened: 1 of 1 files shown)


mdietze:
First, this change appears to result in the creation of a stray unused parameter, soilWFracInit, which is still defined and read in but never used. This parameter is in all existing input files and input workflows, making this a non-backwards compatible breaking change.

Second, the change itself seems to be fairly trivial (initializing soil moisture in absolute terms rather than relative), and in most cases will be undesirable. The ways most people think about and measure soil moisture are on a relative basis (% moisture or % saturation) not on an absolute basis (total mm of water in the soil column). Most of the data to initialize soil moisture will likewise be %. Absolute soil moisture is also very sensitive to structural choices about the assumed rooting depth at a site -- the current approach to initialization acknowledges this by allowing users to specify the (fairly constrained) % moisture and the (fairly unconstrained) WHC separately. In most cases this makes the most sense. There are inevitably cases where the pipelines feeding SIPNET will need to multiply or divide by WHC as part of setting the IC, but we want that to be the edge case not the normal case. This proposal reverses that norm.

PR #65, opened 2025-04-25
Revamp .clang-tidy 

Alomir:
I was never _really_ happy with our `.clang-tidy` file, as it was just a dump of CLion's config of dubious provenance (with some small exclusions).

This PR changes the strategy, making the file a small exclusion list from the clang-tidy default, instead of turning all checks off and then adding back a subset.

There are a pair of NOLINT directives in the code that tell clang-tidy to ignore the line, as they had warnings that were (a) false positives as far as I can tell, and (b) not warnings I want to totally shut off.


dlebauer (approved):
LGTM. Makes sense to punt removal deprecated files & functions to another day.

PR #66, opened 2025-04-28
Add Wishlist Page

Alomir:
Just a simple place to jot down ideas for future work. The intent here is to provide a space to record ideas without needing to flesh them out enough for feature requests.

### DIFF
--- a/docs/wish-list.md
+++ b/docs/wish-list.md
@@ -1,0 +1,74 @@
+# SIPNET WISH LIST
+
+This is the spot to capture ideas on future work that have not been fully thought out (and should thus go into a 
+feature request).
+
+Order is not relevant here, though we _could_ try to order by priority, I suppose. Seems like that will get out of date
+very easily, though. Also, the infrastructure vs Modeling divide is loose at besst.
+
+## Infrastructure Ideas
+
+### More doxygen docs
+We have doxygen set up, and we even have it auto publish, but... we don't have a lot of actual documentation. This is 
+something that could (and should) be incrementally improved as we work on the code, but it might be a really good idea
+spend some focused time on this.
+
+### Remove unsupported code
+There are a lot of compiler options and related code that we have no intention of supporting, and we are not even sure 
+what still works.
+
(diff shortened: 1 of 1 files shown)


infotroph (approved):
Thanks for starting this! Lots of good ideas here.

docs/wish-list.md
```diff
@@ -0,0 +1,74 @@
...
+We might also want to have a verbosity option to control how much gets reported.
+
+## Modeling Ideas
+
+### TBD
+
+## Lesser Ideas (in either scope or importance)
+
+### Have events work with ROOTS=0 (unless we remove that case!)
+The event handler requires `ROOTS` to be on; this is apparent in event modeling that
```
dlebauer, line 71:
```suggestion
[Only if motivated by specific use case] The event handler requires `ROOTS` to be on; this is apparent in event modeling that 
```

docs/wish-list.md
```diff
@@ -0,0 +1,74 @@
...
+* configuration options in effect
+* version of SIPNET
+* warnings and errors (possibly all already there?)
+* information flow description
+
+We might also want to have a verbosity option to control how much gets reported.
+
+## Modeling Ideas
+
+### TBD
```
dlebauer, line 66:
```suggestion
### Features excluded in favor of simplicity

May be revisited if motivated by specific use cases or to improve model skill

If to improve model skill, improvements should be demonstrated under multiple conditions.

* Multi-layer soil  
* Rooting zone differences by crop type  
* Sub-daily irrigation  
* Represent subdaily irrigation as duration as well as amount.  
* Additional Q10   
* Split mineral N into NO3 and NH4 and explicitly represent nitrification and denitrification  
* CH4 flux: account for diffusion, methane oxidation, ebullition, and plant transport  
* Methanotrophy  
* N immobillization  
* Variable Plant CN Ratios  
* Water holding capacity of SOM  
* Fruit, nut, seed pools  
* Orchard floor preparation practices  
* Clay / Sand  
* Lignin  
* pH

```

PR #68, opened 2025-04-30
Remove TOC from parameters.md

dlebauer:
Mkdocs auto-generates TOC, so this manually curated TOC is redundant (and prone to becoming out of date!)

image


Alomir (approved):
Looks good!

PR #76, opened 2025-05-26
Remove Unsupported SIPNET Modes

Alomir:
This PR:
* Removes unsupported SIPNET modes senstest and montecarlo
* Removes code associated with those modes
* Updates sipnet.in for this change
* Checks the RUNTYPE mode is sipnet.in for old mode choices and errors with message if found


dlebauer (changes requested):
@Alomir thanks for this PR and nice work removing obsolete code. I think we could remove more!

Is it time to bump to v2.0?

TODO:

Not clear which of the following need to be part of this PR vs documented in a new issue, but  

- Update changelog 
  - Breaking: removed obsolete run types senstest and estimate and associated code
  - Deprecated: "RUNTYPE" is obsolete. Will be silently ignored if set to 'standard'. Runs in 'standard' mode by default. 
- remove refs to estimate and senstest from 
  - Makefile
  - sipnet.c|h
  - paramchange.c|h
  - README.md
  - delete entire `estimate/` directory
  - remove bintotxt & txttobin utilities; these are associated w/ estimate mode 
  - See also grep or :
    - https://github.com/PecanProject/sipnet/search?q=estimate # some results unrelated
    - https://github.com/PecanProject/sipnet/search?q=senstest

Nice to have:
- With removal of Senstest and Estimate, it seems the `*.param` files should only require the first two cols (`parameterName value`) whereas the last four (`Estimated minValue maxValue Sigma` are obsolete. Are both valid formats? I'd propose we either allow both formats but send a message like `ignoring unused columns 3‑6 in <paramfilename>`.

Alomir:
Thanks for the great review, @dlebauer! I think all of your items should fit either into this PR, or in a "part 2" if we want to keep them smaller. Once all of that list is done, I think it would be a fine time for v2.0

PR #77, opened 2025-05-26
Fixes event file order check issues

Alomir:
This PR:
- fixes errors in the event file reader function that checks to make sure the events are in order
- adds test points to test changes and lock down behavior


infotroph (approved):
Looks great! Tested and working correctly on the file that inspired the bug report 👍

PR #100, opened 2025-06-13
Add edit pencil to docs and change color from indigo to green

dlebauer:
PR #97 from a few days ago was insufficient to add edit pencil to the page b/c we aren't using default theme. 

now, see pencil:

image


Alomir (approved):
Looks good!

PR #109, opened 2025-07-03
SIP103 Revamp Smoke Tests

Alomir:

PR #110, opened 2025-07-03
Smoke test revamp draft

Alomir:
This is just a temp place to dump some changes, should not be merged in. See #109 for the real thing.


Alomir:
No longer needed

PR #114, opened 2025-07-07
SIP78 Convert switches to run time options part 3

Alomir:
This PR:
- converts all remaining compile-time switches to run-time options
- updates help text
- adds validation check for conflicting option settings
- enables all smoke tests
- removes additional dead code from sipnet.c
- adds `soilBreakdown()` required by `LITTER_POOL` 
- updates `src/sipnet/sipnet.in` to show all possible options in that file
- updates cli-options.md
- updates CHANGELOG.md

Most of the changed lines are from replacing two of the smoke test expected output files. Files of interest for reviewing:
- `docs/*`
- `src/*`
- `tests/smoke/run_smoke.sh`
- `tests/smoke/*/sipnet.in`
 
Note: in that "removes additional dead code" bucket, the following variables/params/functions got trimmed. The TrackerVars get set but are never accessed outside of that. I suspect that these were used by functions/modes that themselves have been removed. I did not remove obsolete `params`, as I worry that will cause I/O problems with sipnet clients; they have been marked as obsolete in the code.
- `TrackerVars` `fa`, `fr`, `plantWoodC`, and `LAI`
- functions `simpleWaterFlow`, `transSoilDrainage`

Issue #78, opened 2025-05-27
Convert SIPNET compile-time flags to run-time options


Alomir:
Once [Decide on fate of compiler switches](https://github.com/PecanProject/sipnet/issues/71) is complete, update SIPNET:
- Flags slated for removal: remove flag and all related code
- Flags slated for "always on": remove flag and all "flag off" code branches
- Flags slated to stay as options: convert to run-time options

For the flags that are being converted, 
- expand the existing command line option code to include them
- add checks for exclusions and dependencies between options
- expand `sipnet.in` handling to also handle the same set of options as config-file params; note that specifications at the command line override those specified via file
- (likely) add a context struct or the like to track all of the config that is specified; possibly add source of specification (file vs command) in there. We will eventually want to log that info.

See issue https://github.com/PecanProject/sipnet/issues/72 for discussion of current compile flag exclusions and dependencies (but note that list may not be exhaustive).

PR #96 Covers the first part of the above, but is not enough to close this issue.

dlebauer:
We've reached consensus in https://github.com/PecanProject/sipnet/issues/71#issuecomment-2935967524

PR #123, opened 2025-07-15
Update docs/parameters.md to reflect deprecated and obsolete params

dlebauer:
This PR updates docs/parameters.md to align with the SIPNET implementation as it will be following completion of #78 by merging #114. 

- removal of obsolete parameters
- annotate unused soilWetness from clim file
- replace compile-time parameters with "Run-time options" and "Hard coded values"

More work is likely needed - I am just adding these as I review #114.

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -281,7 +281,27 @@ Run-time parameters can change from one run to the next, or when the model is st
 
 ## Run-time Options
 
-This section replaces the previous "Compile-time parameters" section, as most of these are now configurable at run-time.
+Configuration settings are applied in the following order of precedence:
+
+1. Default values built into SIPNET
+2. Values from the configuration file
+3. Command-line arguments
+
+Thus, command-line arguments override settings in the configuration file, and configuration file settings override default values.
+
+### Input / Output Flags
+
+| Option                  | Default   | Description                                                                 |
+|-------------------------|-----------|-----------------------------------------------------------------------------|
+| `input-file`            | sipnet.in | Name of input config file                                                   |
+| `file-name`             | sipnet    | Prefix of climate and parameter files                                       |
+| `do-main-output`        | 1         | Print time series of all output variables to `<file-name>.out`              |
+| `do-single-outputs`     | 0         | Print outputs one variable per file (e.g. `<file-name>.NEE`)                |
+| `dump-config`           | 0         | Print final config to `<file-name>.config`                                  |
+| `print-header`          | 1         | Whether to print header row in output files                                 |
+| `quiet`                 | 0         | Suppress info and warning message                                           |
+
+### Model Flags
 
 | Option                  | Default | Description                                                              |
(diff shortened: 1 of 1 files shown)


docs/parameters.md
```diff
@@ -287,48 +279,123 @@ Run-time parameters can change from one run to the next, or when the model is st
...
+| `growth-resp`           | 0       | Explicitly model growth respiration, rather than including with maintenance respiration. |
+| `leaf-water`            | 0       | Calculate leaf pool and evaporate from that pool.                        |
+| `litter-pool`           | 0       | Enable litter pool in addition to single soil carbon pool.               |
+| `microbes`              | 0       | Enable microbe modeling.                                                 |
+| `snow`                  | 1       | Keep track of snowpack, rather than assuming all precipitation is liquid.|
+| `soil-phenol`           | 0       | Use soil temperature to determine leaf growth.                           |
+| `soil-quality`          | 0       | Use soil quality submodel.                                               |
+| `water-hresp`           | 1       | Whether soil moisture affects heterotrophic respiration.                 |
+| `num-carbon-soil-pools` | 1       | Number of carbon soil pools.                                             |
+| `do-main-output`        | 1       | Print time series of all output variables to `<file-name>.out`.          |
```
dlebauer, line 319: Need to delete duplicate rows

docs/parameters.md
```diff
@@ -287,48 +279,123 @@ Run-time parameters can change from one run to the next, or when the model is st
...
+
+#### Example Configuration File
+
+```
+# Base filename (used for derived filenames if not explicitly specified)
+file-name = mysite
+
+# Input files
+param-file = mysite.param
+clim-file = mysite.clim
```
Alomir, line 359: These are not settable in the config file (or via command line). They'll get read, but they will be overridden by calculation from `file-name`

docs/parameters.md
```diff
@@ -287,48 +279,123 @@ Run-time parameters can change from one run to the next, or when the model is st
...
+SIPNET reads a configuration file that specifies run-time options without using command-line arguments. By default, SIPNET looks for a file named `sipnet.in` in the current directory. These will be overwritten by command-line arguments if specified.
+
+The configuration file uses a simple key-value format, `option = value`. 
+With one option per line; comments follow `#`.
+
+#### Example Configuration File
+
+```
+# Base filename (used for derived filenames if not explicitly specified)
+file-name = mysite
```
Alomir, line 355: I definitely like this form ('snake-case', as it's called), given the parallel with the command line - but it's a deviation from all of the `sipnet.in` files in the codebase (which use ALL_CAPS versions of these params). Do we care?

infotroph: VERY IMPORTANT correction: `snake_case` uses underscores. This is obviously `kebab-case` 😛

But more seriously I vote we enforce identical spelling on command line and sipnet.in so that we don't have to remember differences like underscore vs hyphen.

PR #125, opened 2025-07-16
SIP116 Remove SOIL_QUALITY mode

Alomir:
This PR:

* Removes soil-quality mode from SIPNET
* Updates smoke test config output to match
* Updates docs to match

NOTE! It really feels like we have a bug in the multipool case - `microbeGrowth()` and `soilDegradation()` both have `if (ctx.soilMultipool)` conditionals that are doing nothing, and there are comments that indicate there should be something. This PR doesn't change this either way, just makes it clearer.

UPDATE: we are removing multipool support as well; note that this is a bigger change.


dlebauer (changes requested):
Thanks for this PR, and for digging in to extract the SOIL_QUALITY logic — that surfaced a point that I hadn’t fully appreciated: soil quality is what drives the multipool dynamics, without it there are no decomposition dynamics for the multi-pool case.

I’d recommend that if we’re deprecating SOIL_QUALITY, we also deprecate the multipool option. It simplifies the code and model logic, and will make things cleaner as we move toward adding an N cycle.

Alomir:
Soil multipool support has been removed

PR #127, opened 2025-07-16
SIP126 Remove soilWetness

Alomir:
This PR:

* Removes `soilWetness` from sipnet's `ClimateVars`
* Updates the climate file read function to handle the removal
* Updates smoke tests so that both "legacy" and "new" file formats are tested
* Updates CHANGELOG

After this change, sipnet will understand two climate file formats:
* Legacy format, with 14 total columns including both location and soilWetness
* Current format, with 12 columns
Climate files with one of those two columns removed will cause an error, as it is very difficult for sipnet to know which one.


Alomir:
@dlebauer Done! Good catch.

infotroph (approved):

PR #130, opened 2025-07-19
tried to capture ideas for re-implementing multipool

dlebauer:
I just wanted to add enough breadcrumbs to capture some thoughts from how multipool might be added in the future.

Please correct or add anything else that might be useful


Alomir:
Great idea to capture this. 

Can we also have harvest/fert/tillage events apply to the top layer, and lket the lower payer affects be calculated from C:N or some other function? I'd love to avoid needing different event specs (i.e., more params) based on number of soil layers.

Alomir (approved):
Looks good!

PR #131, opened 2025-07-21
SIP128 Add fert event handling (C only)

Alomir:
This PR:
* Adds basic fertilization event handling (just C, no N)
* Updates harvest handling for the no-litter-pool case
* Updates tests for harvest change
* Updates output formatting for unit tests
* Adds test for basic fert event


dlebauer (commented):
Looks great! Good to go once documentation in #135 is incorporated.

PR #135, opened 2025-07-31
Update model structure docs to explain what happens when LITTER_POOL is off

dlebauer:
### Key changes

1. Added the `### Litter Pool` section that attempts to provide a high level overview of what happens when litter is off. The goal was to make this very general, prioritizing the full model (LITTER_POOL=1) case since reflecting if / else cases throughout the documentation could make it more confusing than not. _Is this the right place to put it? Is it correct and sufficient to be correct?_
2. taking the `\frak{}` formatting to indicate that OM addition has been added.

### small changes
 
- mostly, cutting lines to ~100 chars b/c that was quicker than figuring out why word wrap wasn't working!? 
- And a few word changes for clarity, fixing citation


Alomir (approved):
This looks good. All of my questions happen when litter pool is ON, not off - but that's a separate discussion!

PR #139, opened 2025-08-15
SIP136 Fix LITTER_POOL lack of rSoil calc

### LINKED ISSUE #136: Investigate lack of rSoil flux update
In `sipnet.c:soilDegradation()`, when LITTER_POOL is on, it seems that `fluxes.rSoil` is never being updated. it is being used to modify `envi.soil`, as expected, but as the flux itself is never being calculated, this is unlikely what was intended.

Probable fix: as is done in the non-MICROBE/non-LITTER POOL case at the bottom, we might just be missing:
```
fluxes.rSoil = fluxes.maintRespiration`
```
at the top of that code block.

Alomir:
fixes #136 

This PR adds proper setting of `fluxes.rSoil` in the LITTER_POOL code block section of the `soilDegradation` function

Also, this PR trims the smoke test inputs to two years of data, since we were going to be changing some of the outputs with this fix.


dlebauer (approved):
Good catch, that could have caused a lot of headache down the road!

Issue #84, opened 2025-06-03
Remove transpose and subset utilities


dlebauer:
This is low priority, but also likely not terribly difficult. Feel free to close as 'not planned'; I just wanted to write this down while I was thinking about it. This could also be left as a 'good first issue'.

As we are cleaning and refactoring code, I think that the subset.c and transpose.c files can be safely removed. This is exactly the type of work that is easier (and more reliable) to do in a higher level language like R or Python than using a hand-rolled implementation. 

These files have 100s of lines of code to do what R can do in a few lines, e.g. transpose.c can be replicated in R with `x <- read_table(file); t(x); write_table(x, file)`.

dlebauer:
So does this come down to 

```
git rm -rf src/utilities
```

?

Alomir:
And as such, this has become so simple that I went ahead and spent the 5 minutes required to do it.

PR #141, opened 2025-08-17
SIP84 Remove utilities

Alomir:
This PR removes the transpose and subsetData utilities, and cleans up the Makefile for their removal.


dlebauer (approved):
Always nice to shed unused code!

fixes #84

PR #146, opened 2025-08-22
SIP124 Remove obsolete params

### LINKED ISSUE #124: Clean up obsolete params in sipnet.c
UPDATE: this issue now deals only with the OBSOLETE params; `soilWetness` has been split off into #126.

After all of our work removing unused functionality in sipnet, there are several parameters that have been made obsolete. All but one of these are in the `Parameters` struct, and are labeled as `OBSOLETE`.

In general, we want to remove these from `sipnet.c`, as well as any lingering references elsewhere in the code (there should be no _uses_ of these variables, though some of them may still be getting assigned in places).

For these parameters, we should update the code that reads in the `*.param` file so that it ignores unknown params, and throws a warning similar to what we do for `sipnet.in` - that is, issue one warning line that lists all the unknown params that were encountered. No need to litter the output with multiple lines.

Alomir:
Fixes #124

This PR:
* Removes obsolete params
* Removes old handling of params marked as obsolete
* Changes reading of params to warn on unknown params instead of error
* Updates testParamInput


dlebauer (approved):
🥳

PR #148, opened 2025-08-25
SIP147 Retool events as fluxes

Alomir:
Note: intial retool as fluxes is in the first commit, which shows that `sipnet.out` has not changed (no green check, though, as `events.out` has)


dlebauer (approved):
Looks nice - I like the change to flux-based event handler and the general direction of ordering flux calcs then pool updates! And thanks for catching the bug in how immedEvap was handled.

- Since I only see changes to the irrig flux, I assume that this PR doesn't actually change the values or implied units written out, it only changes how they are internally calculated. Correct? 
- see two inline comments:
  - check for climLen <=0?
  - change comment from (update pools --> update fluxes).

Alomir:
> Since I only see changes to the irrig flux, I assume that this PR doesn't actually change the values or implied units written out, it only changes how they are internally calculated. Correct?

@dlebauer : I think this IS changing the implied units, from pool amounts (eg g C/m^2) to fluxes (eg g C/m^2/day)

dlebauer:
@Alomir 

> I think this IS changing the implied units, from pool amounts (eg g C/m^2) to fluxes (eg g C/m^2/day)

What I meant was probably unclear but it was just that when units were previously mass/area, the daily time step implicitly added the /d even if this wasn't explicit. Which is why the outputs don't change.

PR #152, opened 2025-08-28
Update documentation

dlebauer:


docs/model-inputs.md
```diff
@@ -0,0 +1,289 @@
...
+- notes: PFT (crop type) is not an input parameter for a planting event because SIPNET only represents a single PFT.
+
+### Harvest
+
+| parameter                                                  |  col  | req?  | description           |
+| ---------------------------------------------------------- | :---: | :---: | --------------------- |
+| fraction of aboveground biomass removed                    |   5   |   Y   |                       |
+| fraction of belowground biomass removed                    |   6   |   N   | default = 0           |
+| fraction of aboveground biomass transferred to litter pool |   7   |   N   | default = 1 - removed |
+| fraction of belowground biomass transferred to litter pool |   8   |   N   | default = 1 - removed |
```
infotroph, line 179: If I'm interpreting the current code right these are all required at the moment. Is there a plan to change this?

docs/user-guide/model-inputs.md
```diff
@@ -0,0 +1,289 @@
...
+- notes: PFT (crop type) is not an input parameter for a planting event because SIPNET only represents a single PFT.
+
+### Harvest
+
+| parameter                                                  |  col  | req?  | description           |
+| ---------------------------------------------------------- | :---: | :---: | --------------------- |
+| fraction of aboveground biomass removed                    |   5   |   Y   |                       |
+| fraction of belowground biomass removed                    |   6   |   N   | default = 0           |
+| fraction of aboveground biomass transferred to litter pool |   7   |   N   | default = 1 - removed |
+| fraction of belowground biomass transferred to litter pool |   8   |   N   | default = 1 - removed |
```
dlebauer: @infotroph good catch, and yes it appears that these are all required in the current implementation. These defaults make sense for annuals but not perennials, so I think it should stay this way for simplicity.


```suggestion
| fraction of belowground biomass removed                    |   6   |   Y   | |
| fraction of aboveground biomass transferred to litter pool |   7   |   Y   | |
| fraction of belowground biomass transferred to litter pool |   8   |   Y   | |
```

@Alomir leaving this for you to make sure we are on the same page about the updated requirements.

dlebauer: FYI the requirement is enforced by https://github.com/PecanProject/sipnet/blob/85a136bf6078b2b2943d92b06a577fde15cc0844/src/sipnet/events.c#L47-L49

PR #161, opened 2025-09-25
Add woodCreation as output

cami-webb:
Added woodCreation back as an output of sub.sipnet.output

--- 

@dlebauer edit:

PR motivation (from separate Slack thread): the intent of writing woodCreation to sipnet.out is to use it as an estimate of Gross Woody Biomass Increment ([GWBI](https://github.com/PecanProject/pecan/blob/f548e761d9f10cdb1162feeb985046ec28657d49/base/utils/data/standard_vars.csv#L35)), which will be used for assimilation of tree ring and forest inventory data.

### DIFF
--- a/src/sipnet/state.h
+++ b/src/sipnet/state.h
@@ -577,7 +577,7 @@ typedef struct TrackerVars {  // variables to track various things
   // g C * m^-2 litterfall, year to date: SUM litter
   double yearlyLitter;
 
-  // g C * m^-2 wood creation 
+  // g C * m^-2 wood creation
   double woodCreation;
 
 } Trackers;


src/sipnet/sipnet.c
```diff
@@ -986,7 +986,7 @@ void calcRootResp(double *rootResp, double respQ10, double baseRate,
 // This is in addition to [1], source not yet determined (controlled by
 // ctx.growthResp)
 void vegResp2(double *folResp, double *woodResp, double *growthResp,
-              double baseFolResp) {
+              double baseFolResp, double gpp) {
```
Alomir:
```suggestion
              double baseFolResp) {
```
I think this was an incorrect merge, this arg should have gone away

PR #171, opened 2025-10-17
SIP86 Add soil mineral nitrogen pool and nitrogen cli option

### LINKED ISSUE #86: Add soil mineral‑nitrogen state variable (`N_min`)
- [x] Add runtime flag to turn N Cycle on / off (default: off)
- [x] Add new state variable N_min
- [x] Read in N_(min,0) from parameter file
- [x] Write N_min to sipnet.out at each time step
- [x] Update documentation
- [x] write test to ensure that specified value of N_(min,0) is written to sipnet.out

Alomir:
Fixes #86 

This PR:
- adds a new `envi` variable, `minN`, to track the soil mineral nitrogen pool
- adds a new parameter, `minNInit`, to initial that pool
- writes `minN` to `sipnet.out` in a new column labeled `minN`
- adds a new run-time option, `nitrogen-cycle`, that controls whether `minN` is used/updated.
- re-purposes the `russell_2` smoke test as a new nitrogen-cycle smoke test to verify that the above works
- updates outputs of the other smoke tests (I verified that the new column is the only change)
- updates `model-inputs.md` and `model-outputs.md`

### DIFF
--- a/docs/model-structure.md
+++ b/docs/model-structure.md
@@ -33,33 +33,25 @@ Model state is updated in the following order:
 2. Process events — convert events to per‑day fluxes and accumulate into fluxes.
 3. Update pools — pools are updated from the accumulated fluxes and pool‑specific updates.
 
-### Notes on notation:
-
-Fluxes are denoted by $F$, except that respiration is denoted by $R$ following convention and previous descriptions of SIPNET.
-
-Parameters and other information can be found in the Parameters documentation. <!-- or maybe combined here?-->
 
 ### Scope
 
 This document provides an overview of the SIPNET model’s structure. It was written to 
+
 - Consolidate the descriptions from multiple papers (notably Braswell et al 2005 and Zobitz et al 2008). 
 - Provide enough detail to support the addition of agronomic events, CH4, and N2O fluxes.
 - Focus on features currently in regular use.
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/model-inputs.md
```diff
@@ -224,20 +224,22 @@ Thus, command-line arguments override settings in the configuration file, and co
...
+| `litter-pool`    | off     | Enable litter pool in addition to single soil carbon pool                               |
+| `microbes`       | off     | Enable microbe modeling                                                                 |
+| `nitrogen-cycle` | off     | Enable nitrogen cycle modeling                                                          |
+| `snow`           | on      | Keep track of snowpack, rather than assuming all precipitation is liquid                |
+| `soil-phenol`    | off     | Use soil temperature to determine leaf growth                                           |
+| `water-hresp`    | on      | Whether soil moisture affects heterotrophic respiration                                 |
 
 Note the following restrictions on these options:
  - `soil-phenol` and `gdd` may not both be turned on
+ - `events` and `microbes` may not both be turned on
```
dlebauer: Should we also prohibit nitrogen + microbes both on? I suspect it would be complicated and at this point unnecessary to support the combination.

PR #186, opened 2025-11-13
Clarify soilWetnessFrac calc in documentation

dlebauer:
Updated the description of soilWetnessFrac to clarify it is a mean

Adding this note so that when someone considers the relationship between soilWater and soilWetnessFrac, they will understand why the ratio varies, i.e. soilWater/soilWetnessFrac != WHC

Additional options considered:
- change name to meanSoilWetnessFrac? not done b/c low importance
- report current soilWetness? not done for backward compatibility.


Alomir (approved):
Looks good!

PR #190, opened 2025-11-18
Add GitHub Actions workflow for docs deployment

dlebauer:
old build system no longer working, I think because settings were changed from old process that used gh-pages to Github's new github-pages environment.


dlebauer:
Bypassing PR review b/c I just want to see if I can get the pages to publish. The deploy pipeline was broken.

PR #191, opened 2025-11-18
Consolidate docs deployment workflows

dlebauer:
remove deprecated deploy.yml add PR preview to publish-docs.yml


infotroph:
As far as I can tell that "view deployment" button goes to the live main site (https://pecanproject.github.io/sipnet/) and not to a PR-specific URL. Will it behave differently in PRs that contain doc changes?

PR #192, opened 2025-11-18
Require PR approval by specific people

dlebauer:
Now that we are getting more external contributors we can require approvals by specific people using this codeowners file.

Before or after merging we should set "Require review from Code Owners" on this page https://github.com/PecanProject/sipnet/settings/rules/1432057

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # All files in the repository must be reviewed by one of these users
-*   @Alomir @dlebauer @Dietze
+*   @Alomir @dlebauer @mdietze @infotroph @mswilburn 


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# All files in the repository must be reviewed by one of these users
+*   @Alomir @dlebauer @Dietze
```
Alomir: Github claims this is an error - @Dietze is an unknown user

dlebauer:
```suggestion
*   @Alomir @dlebauer @mdietze @infotroph @mswilburn 
```

dlebauer:
It lets us separate write access from PR-approval requirements.

PR #205, opened 2025-12-13
Update documentation of C:N calls

dlebauer:
@alomir please update the description if needed.

In particular this sentence should reflect that CN is calculated on demand when used (if that is correct, as discussed in recent conversation about creating a cn calculating function instead of adding a state variable).

```
In SIPNET, the C:N ratio of soil and litter pools is calculated from carbon and nitrogen pools and is used to calculate C:N-dependency $D_{CN}$ in Eq. \eqref{eq:cn_dep}.
```

---
_note:_ there are many commits assoc. w/ this PR b/c I did `git merge --no-ff` instead of rebase. I was just trying minimize merge drama; this is apparently expected behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ $(SIPNET_LIB): $(SIPNET_OFILES)
 GCC_VERSION = $(shell $(CC) --version)
 info:
 	@echo "System info"
-	@echo "ARCH: $(shell arch)"
+	@echo "ARCH: $(shell uname -m)"
 	@echo "CC: $(GCC_VERSION)"
 	@echo ""
 


docs/model-structure.md
```diff
@@ -290,10 +290,10 @@ $$\small j \in \{\text{soil, litter}\}$$
 To represent the influence of substrate quality on decomposition rate, we add a simple dependence function $D_{CN}$.
 
 $$
-  D_{CN} = \frac{1}{1+k_CN \cdot CN} \tag{11}\label{eq:cn_dep}
+  D_{CN} = \frac{1}{1+k_{CN} \cdot CN} \tag{11}\label{eq:cn_dep}
 $$
 
-Where $k_CN$ is a scaling parameter that controls the sensitivity of decomposition rate to C:N ratio. This parameter represents the half-saturation constant of the Michaelis-Menten equation.
+Where $k_{CN}$ is a scaling parameter that controls the sensitivity of decomposition rate to C:N ratio. This parameter represents the half-saturation constant of the Michaelis-Menten equation.
```
Alomir: Hmmm, this is not really a Michaelis-Menten form, so that much of this comment should probably change.
 
Do we want to formulate D_{CN} as (k/(k+CN), and label k as "the CN ratio at which the dependency is 1/2"? (This is identical to the current form if k_new is equal to 1/k_current)

We could also formulate this in an analogous way with the D_N_min equation, and use the form 1/(1+CN/k), which is equivalent to k/(k+CN).

PR #203, opened 2025-12-10
Fix and standardize parameters.md — notation, tables, units, and row numbering

ayushman1210:
##Summary:

- Cleaned and standardized [parameters.md]
- Fixed notation, LaTeX unit formatting, and inconsistent subscripts/superscripts.
- Consolidated and renumbered run-time parameter tables (rows 1–79).
- Added missing parameter symbols and parameter names; clarified definitions and notes.

##What I changed (high level):

- **Notation**
- Added/matched flux symbols (Rh, Ra, ET, T) and parameter symbol SLW.
- Added superscript exception note and X_HC subscript definition.
- **Tables and formatting**
- Converted mixed/broken sections to consistent Markdown long tables.
- Standardized use of \text{...} in LaTeX math and fixed malformed subscripts (e.g., K_{VPD,exp}, Q_{10,\text{soil}}).
- Fixed units to LaTeX math format throughout.
- **Run-time parameters**
- Sequential Row column added and numbered 1..79 for all run-time parameters.
- Assigned parameter configuration names (camelCase) where missing.
- Consolidated autotrophic/heterotrophic respiration, photosynthesis, litter, allocation, nitrogen, methane, water, and tree physiology sections.
-**Minor**
- Clarified notes on units and internal conversions (e.g., year^-1 → per-day conversion).
- Kept Hard-coded Values section unchanged except formatting consistency.

fixes #81

PR #225, opened 2026-01-17
SIP211 context metadata cleanup

### LINKED ISSUE #211: Issue: Potential Memory Leak in Configuration Context
***DESCRIPTION***
The configuration context module (src/common/context.c) currently lacks a cleanup mechanism to deallocate memory used for model parameter metadata. 
During model initialization, createContextMetadata allocates memory for struct context_metadata using malloc to populate the ctx.metaMap hash table. However, there is no corresponding function to iterate through the hash table and free these allocations before the program terminates or when a model run is completed.
While the operating system reclaims memory upon process exit, this missing cleanup is a violation of memory management best practices. It poses a risk for future development if SIPNET is integrated into long-running processes, such as optimization loops or library calls, where repeated initializations would cause linear memory growth.

***Solution***
A cleanup function should be implemented to safely destroy the hash table and free all associated metadata structures.

***I am working on this issue thanks***
thanks !!

ayushman1210:
1) Added function declaration in  context.h
- Added destroyContextMetadata() function declaration with documentation explaining its purpose

2) Implemented cleanup function in context.c
Created destroyContextMetadata() that
- Iterates through the ctx.metaMap hash table using HASH_ITER macro
- Safely deletes each entry using HASH_DEL
- Frees the allocated memory for each metadata structure
- Sets ctx.metaMap to NULL after cleanup

3) Integrated cleanup into sipnet.c
Added call to destroyContextMetadata() at the end of cleanupModel() function

Implementation Details
The implementation uses the uthash library's HASH_ITER and HASH_DEL macros for safe iteration and deletion:
- HASH_ITER safely iterates through the hash table while allowing deletion of elements
- HASH_DEL removes each element from the hash table
- free() deallocates the memory for each context_metadata structure
- Setting  ctx.metaMap = NULL prevents potential use-after-free issues

### Fixes
#211


dlebauer (changes requested):
Hi @ayushman1210 thank you for reporting this bug and submitting these changes.

Unfortunately, it looks like these changes. Include your updates to documentation, maybe the same as are in #203.

Could you please either take those files out of this PR or submit a new one so that the PR is focused on the related issue? Either approach would be fine.

And as a minor request, for future reference - it is helpful if you can prepend the **branch name** and **pr title** with the feature number.

dlebauer:
Thanks @ayushman1210 - I'll defer to @Alomir for approval.

src/common/context.h
```diff
@@ -93,6 +93,14 @@ void createContextMetadata(const char *name, const char *printName,
                            context_source_t source, context_type_t type,
                            void *value, int isFlag);
 
+/*!
+ * Cleanup context metadata
+ *
+ * Free all allocated memory in the context metadata hash table.
+ * Call this when done using the context to prevent memory leaks.
+ */
+void destroyContextMetadata(void);
```
Alomir: For naming consistency, this should be better:
```suggestion
/*! 
Free all context metadata allocated during initContext()
Iterates through ctx.metaMap hash table and frees all allocated
context_metadata structures.  Should be called during program
cleanup.
 */
void freeContextMetadata(void);
```

src/common/context.c
```diff
@@ -254,3 +254,12 @@ void printConfig(FILE *outFile) {
     }
   }
 }
+
+void destroyContextMetadata(void) {
+  struct context_metadata *s, *tmp;
+  HASH_ITER(hh, ctx.metaMap, s, tmp) {
+    HASH_DEL(ctx.metaMap, s);
+    free(s);
```
Alomir: I'm not quite sure what cpp-linter is complaining about here, but one way to work around it might be to 
- iterate over the hash and delete the structs
- use `HASH_CLEAR` once that is done

ayushman1210: @alomir Good catch  I’ll refactor this to explicitly free each entry before calling HASH_CLEAR so clang is satisfied.

ayushman1210:
Hey @Alomir Whenever you have some free time, could you please review this PR?
Thanks!!

Alomir (approved):
Looks good!

PR #245, opened 2026-02-01
fixed: Buffer Overflow Vulnerability in countFields Function

ayushman1210:
### Buffer Overflow Vulnerability in countFields() Function

**Location:** `src/common/util.c`, lines 51-62
**Description:**
The `countFields()` function has a fixed-size buffer of 256 bytes but uses `strcpy()` without bounds checking
```c
int countFields(const char *line, const char *sep) {
  // strtok modifies string, so we need a copy
  char lineCopy[256];
  strcpy(lineCopy, line);  // UNSAFE - No bounds checking!
  // ... rest of function
}
```
If the input  line parameter exceeds 255 characters (plus null terminator), `strcpy()` will write beyond the buffer boundary, causing a buffer overflow. This can lead to:
- Stack corruption
- Segmentation faults
- Potential security exploits
- Unpredictable program behavior

This function is called from:
- `readClimData()` in `sipnet.c` (line 164)
- `checkParamFormat()` in `modelParams.c` (line 129)
Both functions process user-supplied input files, making this a exploitable vulnerability.

**changes made**
Replace `strcpy()` with `strncpy()` or use dynamic allocation


Alomir:
Sorry @ayushman1210 - I thought you were going to remove the `writeOutputItemHeaders` code from this PR, I was waiting for that.

ayushman1210:
Hi @Alomir,
I’ve removed the writeOutputHeader from the code, and it looks good. Please review when you have some spare time.
I also have a general question: I’m planning to submit a proposal for the summer, and I’m very interested in two projects:

1. Standardizing Model Couplers Across Models
2. Benchmarking and Validation Framework

Would it be acceptable to submit proposals for both projects to the same organization, or does the program prefer only one proposal per contributor?
Thank you for your guidance

PR #246, opened 2026-02-01
Potential overflow fixed

ayushman1210:
**Location:** `src/sipnet/events.c`, line 183

**Description:**
```c
#define EVENT_LINE_SIZE 1024
char line[EVENT_LINE_SIZE];
```

When reading event lines, the buffer is fixed at 1024 bytes, but there's no check to ensure event parameter strings don't exceed this size.

**Issue:**
While `fgets()` will prevent buffer overflow, extremely long event parameter lines will be silently truncated, potentially causing:
- Incorrect parsing of event data
- Missing event parameters
- Silent data corruption

**Impact:**
- Events with very detailed parameters might be incorrectly processed
- Difficult to debug because no error is reported

**Changes**
Add validation after reading
```c
if (fgets(line, EVENT_LINE_SIZE, in) != NULL) {
  size_t len = strlen(line);
  if (len == EVENT_LINE_SIZE - 1 && line[len-1] != '\n') {
    logError("Event line too long (exceeds %d chars), data may be truncated\n", 
             EVENT_LINE_SIZE);
    exit(EXIT_CODE_INPUT_FILE_ERROR);
  }
}
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # All files in the repository must be reviewed by one of these users
-*   @Alomir @dlebauer @mdietze @infotroph @mswilburn 
+*   @Alomir @dlebauer @mdietze @mswilburn


Alomir (changes requested):
Looks good, just a small suggestion for change

Alomir (approved):
Looks good, thanks!

PR #250, opened 2026-02-05
addressed the uninitialized variable usage risk in modelParams.c by c…

ayushman1210:
#### Uninitialized Variable Usage Risk
**Location:** `src/common/modelParams.c`
line 151
**Description:**
```c
char unknownParams[2048] = "";
char hasUnknownParams = 0;
```
The buffer `unknownParams` is initialized, but operations like `strcat()` assume proper null-termination. If the initialization is somehow skipped or corrupted, undefined behavior could occur.

**Issue:**
While the code initializes the buffer, defensive programming would add additional safeguards.

**changes Addressed **
Add explicit null-termination before first use:
```c
unknownParams[0] = '\0';  // Ensure null-termination
```


Alomir:
Hi @ayushman1210  - I'm not sure what this PR is adding. Initializing with `{0}` is marginally safer, but the buffer is already zero-initialized in the first element, and `strcat` appends a terminating null if there is room.

I think the bigger risk here is the comparison with the `strlen` sum being greater than 2048, which is a (very unlikely) latent bug: `strlen` does not count the terminating nulls, so if the sum is equal to 2048, there won't be room for the needed null in `unknownParams`. Also, 2048 should be a `#define`'d value.

Would you mind thinking that through, and adding those changes as well if you agree?

PR #260, opened 2026-02-13
[N/A] Add methane output

Alomir:
## Summary

- **What**: Add zero-filled column to sipnet.out named `ch4`
- **Motivation**: Enable pipeline implementation

Unit and smoke tests should be sufficient here


dlebauer (approved):
Thank you!

PR #312, opened 2026-04-07
separate tillage effect on litter respiration from litter-to-soil transfer

divine7022:
## Summary

- **What**: 
tillage multiplier was being applied to all of litter breakdown, which meant it sped up litter-to-soil transfer too. Now it only hits litter respiration. Also adds an optional second param (incorporationFraction) so we can model physical mixing separately.
- **Motivation**: 
reduced-till was showing SOC loss in our scenario runs, which makes no sense. Turned out the tillage multiplier on total litter breakdown was pumping more C into soil pool under conventional till than under no-till;  fix is to only let tillage accelerate the CO2 loss side, and handle the physical mixing as its own thing.

- all existing tests still pass
- testEventTillageIncorp covers incorporation with and without N cycle, litter pool off, and old 1-param format
- testBalance still passes so mass balance is fine
- old events.in files work as is, incorporationFraction defaults to 0

## Reproduction steps

If appropriate, list steps to reproduce the change locally

## Related issues

- Fixes #<issue>

## Checklist

- [ ] Related issues are listed above. [PRs without an approved, related issue may not get reviewed](docs/CONTRIBUTING.md#propose-and-receive-feedback).
- [ ] PR title has the issue number in it ("[#<number>] \<concise description of proposed change>")
- [ ] Tests added/updated for new features (if applicable)
- [ ] `docs/CHANGELOG.md` updated with noteworthy changes
- [ ] Code formatted with `clang-format` (run `git clang-format` if needed)

---


src/sipnet/events.c
```diff
@@ -126,14 +126,22 @@ EventNode *createEventNode(int year, int day, int eventType,
...
       TillageParams *tParams = (TillageParams *)malloc(sizeof(TillageParams));
       int numRead = sscanf(eventParamsStr,  // NOLINT
-                           "%lf", &tillEffect);
-      if (numRead != NUM_TILLAGE_PARAMS) {
+                           "%lf %lf", &tillEffect, &incorpFrac);
+      if (numRead < 1) {
         logError("parsing Tillage params for year %d day %d\n", year, day);
         exit(EXIT_CODE_INPUT_FILE_ERROR);
       }
+      if (incorpFrac < 0.0 || incorpFrac > 1.0) {
```
mdietze, line 137: incorpFrac should be a parameter, not an event

mdietze (changes requested):

src/sipnet/events.c
Alomir: Agreed - plus this messes with event structures more than I like, at least at first look

src/sipnet/events.c
```diff
@@ -548,15 +556,31 @@ void processEvents(void) {
...
+        if (tillParams->incorporationFraction > 0.0 && ctx.litterPool) {
+          litterIncorpC = envi.litterC * tillParams->incorporationFraction;
+          fluxes.eventLitterC -= litterIncorpC / climLen;
+          fluxes.eventSoilC += litterIncorpC / climLen;
+          if (ctx.nitrogenCycle) {
+            litterIncorpN = envi.litterN * tillParams->incorporationFraction;
+            fluxes.eventLitterN -= litterIncorpN / climLen;
+            fluxes.eventSoilOrgN += litterIncorpN / climLen;
+          }
+        }
```
Alomir, line 578: This feels like a large change to how tillage works - is there a discussion of this modeling change anywhere?

Alomir (changes requested):

dlebauer:
@divine7022 I believe there was a misunderstanding when we spoke yesterday. 

When you identified an issue with tillage pumping too much C into SOC, thus having the unrealistic effect of protecting it, I didn't mean that we should add a new parameter.

What I meant by "reduce the flux" was to change the fracLitterRespired which partitions between respiration and transfer to SOC.

Considering that we haven't started formal calibration, we are mostly trying to get sensible answers at this point, that we can later formally calibrate.

This does highlight the urgency of formal calibration. But for now let's manually tune a few knobs.

Perhaps first, tillage intensity. Then, fracLitterRespired.

src/sipnet/events.c
divine7022: I should've flagged this before opening the PR, my bad. 
there are kinda two separate changes bundled in here and I think pushback is about the second one.
the bug fix in sipnet.c : right now tillageEffect multiplies all of litterBreakdown, which means it accelerates litter-to-soil transfer alongside respiration. that's what was giving us higher SOC under conventional till than reduced-till in the scenario runs. the fix just moves tillageEffect off of litterBreakdown and onto rLitter only. no new params, nothing to calibrate

new feature (incorporationFraction in events): idea was to let each tillage event carry its own incorporation fraction, since the physical mixing varies with tillage intensity (same way tillageEffect already varies per event in events.in). but I totally get the scope concern -- it needs calibration data we don't have yet.

would it make sense to split this? keep the bug fix, defer incorporationFraction to follow up? happy to strip it down

FYI, 
> Agreed - plus this messes with event structures more than I like, at least at first look

on event structure side -- it's just an optional second field via sscanf, same pattern as reading 1 or 2 values. old events.in files still work unchanged (incorpFrac defaults to 0). shouldn't break anything downstream.

src/sipnet/events.c
divine7022: key thing is behavioral change, which is fair to flag. the issue we hit was that the existing code applied tillageEffect to ALL of litter breakdown (both the fraction that respires and the fraction that transfers to soil). that meant conventional tillage was actually accelerating C transfer into the soil pool, so when you compared reduced-till to conventional, reduced-till showed less soil C, opposite to what reality/literature says.
the fix separates two things that are mechanistically different (1) tillage speeds up decomposition/respiration of surface litter, and (2) tillage physically mixes litter into soil (the new incorporationFraction, which is an instantaneous mass transfer). these are distinct processes -- DAYCENT treats them separately as CLTEFF vs CULTRA

divine7022:
@dlebauer  I think we're mostly on same page but talking past each other a bit.
there's actually two separate things in this PR and I think the confusion is about which one you're pushing back on:

change 1 (the simple fix in calcLitterFluxes): move tillageEffect off of total litterBreakdown and only apply it to rLitter. no new parameters, no calibration needed. this fixes the structural issue where conventional till pumps more C into soil than no-till -- which happens because tillageEffect multiplies both the respired fraction AND the transfer-to-soil fraction equally. tuning fracLitterRespired can shrink the effect but can't eliminate it (the math works out that conv till always has higher litterToSoil as long as fracLitterRespired < 1.0).

change 2 (incorporationFraction): the new optional event param for physical mixing. I hear you that this is out of scope for now and needs calibration data we don't have. totally fine to drop this part.

would it work to split the PR? keep change 1 (the bug fix, no new params) and defer change 2 (incorporationFraction) until we have calibration data? happy to strip it down

mdietze:
FWIW I think it makes sense to have an incorporationFraction parameter, but I think it should be a parameter not an event and should default to 0

dlebauer:
@divine7022 I see what you mean about the structural issue - the ratio `litterToSoil_tilled:litterToSoil_no_till` is set by the tillage effect, so tuning fracLitterRespired can reduce the magnitude of transfer but not remove it.

This was designed to approximate both the biological and physical effects of tillage. The limitation of this structure is that we have a single soil carbon pool with a slow turnover, so transferring litter to the soil carbon pool effectively protects it.

I can also see why a separate `incorporationFraction` could be useful to represent physical mixing explicitly, separate from the biological effect. That said, my preference would be to defer this for now.

My hesitation is mainly about scope and identifiability. We have several uncalibrated parameters that influence the net tillage response. Including:

- litterBreakdownRate
- baseSoilResp
- fracLitterRespired
- tillage effect

So, for me, the question is not whether there is a more realistic representation, but whether the current representation sufficient to capture patterns in the available data once the parameters are tuned - for example, increasing fracLitterRespired and baseSoilResp and reducing the tillage effect.

At this point my preference is to 

- summarize the two changes as separate issues that can be discussed and prioritized explicitly
- explore whether there is a defensible combination of parameters that generates expected responses to tillage
- start setting up site-level calibration runs

PR #318, opened 2026-04-16
Bump numeric version + build tag into -v output

infotroph:
## Summary

Calling `sipnet -v` on the tagged v2.1.0 Sipnet still reports "SIPNET version 2.0.0 (651d66d201)", suggesting it's time to bump the value of NUMERIC_VERSION in version.h.

We can also get a touch more automated by using `git describe` to build tag names into the commit summary -- see below. 

`make clean && make && ./sipnet -v`

* Tested locally with and without uncommitted changes and at commits that are and are not identical to known tags
* Verified that today's result is reasonable in a freshly cloned repo where nobody without needing a `git pull --tags`. It'd be worth checking again once HEAD has moved past the tag -- I remember there may be fine print about when tags are and are not fetched by default.

## Reproduction steps

Current version:

```
% git checkout v2.1.0
% make clean && make && ./sipnet -v
# [compilation output snipped]
SIPNET version 2.0.0 (651d66d201)
% git checkout HEAD^ 
% make clean && make && ./sipnet -v
# [compilation output snipped]
SIPNET version 2.0.0 (c44f579a86)
```

After this change (I created a local v2.1.1 tag for illustration):

```
% git checkout v2.1.1-pre
% make clean && make && ./sipnet -v
# [compilation output snipped]
SIPNET version 2.1.0 (2.1.1-pre)
% git checkout HEAD^ 
% make clean && make && ./sipnet -v
# [compilation output snipped]
SIPNET version 2.1.0 (v2.1.0-1-g601265e399)
```

... so now even without any change to the hard-coded "version 2.1.0" we can automatically see which one was compiled exactly from the tag vs which was a plain old branch tip.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,6 +1,6 @@
 cff-version: 1.2.0
 title: "SIPNET: Simple Photosynthesis and Evapotranspiration Model"
-version: 2.0.0
+version: 2.1.0
 abstract: >
   SIPNET (Simple Photosynthesis and Evapotranspiration Model) is a process-based
   ecosystem model designed to simulate carbon and water fluxes and (optionally) agricultural management
@@ -56,7 +56,7 @@ keywords:
 preferred-citation:
   type: software
   title: "SIPNET: Simple Photosynthesis and Evapotranspiration Model"
-  version: 2.0.0
+  version: 2.1.0
   url: https://github.com/PecanProject/sipnet
   authors:
     - family-names: Michael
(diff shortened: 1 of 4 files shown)


dlebauer (commented):
That is a great idea to distinguish release from mainline head from current commit. 

Since 2.1.0 has been released, this should be 2.1.1. I'll defer to @Alomir to decide how to handle this.

infotroph:
Question: Is the tag in the describe output sufficient (and reliably enough available at build time) for us to drop the numeric version entirely? Then we don't have to remember to bump it for future releases.

Alomir (approved):
Looks good! Added a couple of other missed updates.

PR #319, opened 2026-04-17
SIP27 Fix release binary build naming that prevented uploads

### LINKED ISSUE #27: Automate SIPNET Binary Builds & Manual Release

dlebauer:
### Problem

The SIPNET v2.1.0 release workflow built both binaries successfully, but failed during artifact upload because the binary name and upload path were different. 

The different names were `sipnet-[version]-[os]` vs `sipnet-[os]-[version]`. Going w/ `sipnet-[os]-[version]` to match prior releases.

### Solution

- Update to .github/workflows/release.yml to match release artifact names and upload paths.

**Other updates while in the rabbit hole**

- instead of releasing binaries with names that include os and version, release tar.gz archives following this naming convention, and contain a binary named `sipnet` 
- build explicitly from `macos-arm64` instead of `macos-latest`
- include architecture (x86_64 or arm64) in tar.gz file name
- explicitly use "vX.Y.Z [RELEASE NAME]" as release title placeholder (was "Release vX.Y.Z"

- Verified the artifact naming/path logic now matches:
  - `sipnet-linux-x86_64-<tag>.tar.gz`
  - `sipnet-macos-arm64-<tag>.tar.gz`
- Built the `v2.1.0` binaries locally from the exact release commit `1bd16b782c9941c98abcb9615e8626f1fd78c309`:
  - macOS arm64 binary
  - Linux x86_64 binary (via local Docker)
- Verified `sipnet --version` reports:
  - `SIPNET version 2.1.0 (1bd16b782c)` for both binaries
- uploaded assets to v2.1.0 release 
 
## Reproduction steps (not done)

- For future releases, push a version tag matching `v*.*.*` to trigger the full release workflow.
- To rebuild binaries for an existing tag, run the `Release` workflow manually with `release_ref=<tag>`. (select that branch in the Actions UI, run the Release workflow, and pass release_ref=v2.1.0)
- Confirm the build jobs upload:
  - `sipnet-macos-<tag>`
  - `sipnet-ubuntu-<tag>`

## Related issues

- Fixes #27

## Checklist

- [x] Related issues are listed above. [PRs without an approved, related issue may not get reviewed](docs/CONTRIBUTING.md#propose-and-receive-feedback).
- [x] PR title has the issue number in it ("[#<number>] <concise description of proposed change>")
- [ ] Tests added/updated for new features (if applicable)
- [ ] `docs/CHANGELOG.md` updated with noteworthy changes
- [ ] Code formatted with `clang-format` (run `git clang-format` if needed)

---


.github/workflows/release.yml
```diff
@@ -61,10 +63,14 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           tag_name: ${{ github.ref_name }}
-          name: Release ${{ github.ref_name }}
+          name: ${{ github.ref_name }} [RELEASE NAME]
```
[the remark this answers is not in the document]
dlebauer: @copilot eli12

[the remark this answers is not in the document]
dlebauer: I thought I did update the description. Didn't I @copilot ?

.github/workflows/release.yml
```diff
@@ -30,21 +30,23 @@ jobs:
         run: |
           tag="${GITHUB_REF_NAME}"
```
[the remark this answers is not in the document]
dlebauer: Plz fix @copilot

AritraDey-Dev (approved):
Thanks for the ping! The changes make sense to me. Just to test, I pushed this patch to my forked repo with a tag, which is successfully triggering the workflows.https://github.com/AritraDey-Dev/sipnet/actions/runs/24609489611

.github/workflows/release.yml
```diff
@@ -5,16 +5,19 @@ on:
...
     strategy:
       matrix:
         include:
           - os: ubuntu-latest
-            artifact_os: ubuntu
+            artifact_name: linux-x86_64
             binary_path: sipnet
-          - os: macos-latest
-            artifact_os: macos
+          - os: macos-arm64
```
AritraDey-Dev: is there any reason to use arm64 here?

AritraDey-Dev: I’m asking this because it takes a good amount of time to get a macOS ARM64 runner,the job gets queued and might even time out. If there’s no specific goal, I think it’s better to stick to the latest version.

dlebauer: The intent was to make the build system explicit rather than "whatever latest points to". But if that is more pedantic than practical, I'm happy to change.

AritraDey-Dev: https://github.com/AritraDey-Dev/sipnet/actions/runs/24609449828 ,so i tested with macos-arm but it was not able to get the runner.Would suggest to stick to latest.

dlebauer:
```suggestion
          - os: macos-latest
```

AritraDey-Dev (approved):

.github/workflows/release.yml
```diff
@@ -25,26 +28,28 @@ jobs:
...
+          tar -C dist -czf "dist/${archive_base}.tar.gz" sipnet
+          rm -f "dist/sipnet"
 
       - name: Upload release artifact
         uses: actions/upload-artifact@v4
         with:
-          name: sipnet-${{ github.ref_name }}-${{ matrix.artifact_os }}
-          path: dist/sipnet-${{ github.ref_name }}-${{ matrix.artifact_os }}
+          name: sipnet-${{ matrix.artifact_name }}-${{ github.ref_name }}
+          path: dist/sipnet-${{ matrix.artifact_name }}-${{ github.ref_name }}.tar.gz
```
Alomir: Why don't we use `artifact-base` here? That way it exactly matches the tar operation, and if we change `artifact-base`'s definition, it automatically updates here too.

.github/workflows/release.yml
```diff
@@ -25,26 +28,28 @@ jobs:
         run: make
         shell: bash
 
-      - name: Prepare release binary
+      - name: Prepare release archive
         shell: bash
         run: |
           tag="${GITHUB_REF_NAME}"
```
Alomir, line 35: Is "${GITHUB_REF_NAME}" the same as ${{ github.ref_name }}?

Alomir (changes requested):
Small suggestion in comments

.github/workflows/release.yml
dlebauer: according to my assistant:

Yes, for this workflow’s tag-triggered release use case, they refer to the same short ref name.

* `${GITHUB_REF_NAME}` is the environment variable form available inside shell steps.
* `${{ github.ref_name }}` is the GitHub Actions expression/context form available in workflow YAML.

So inside `run: |, tag="${GITHUB_REF_NAME}"` is appropriate. Outside shell, such as with: fields, use `${{ github.ref_name }}` unless you pass it through an output like we did.

.github/workflows/release.yml
```diff
@@ -25,26 +28,28 @@ jobs:
         run: make
         shell: bash
 
-      - name: Prepare release binary
+      - name: Prepare release archive
         shell: bash
```
dlebauer, line 33:
```suggestion
        id: prepare_archive
        shell: bash
```

.github/workflows/release.yml
dlebauer:
```suggestion
          name: ${{ steps.prepare_archive.outputs.archive_base }}
          path: dist/${{ steps.prepare_archive.outputs.archive_base }}.tar.gz
```

.github/workflows/release.yml
```diff
@@ -25,26 +28,28 @@ jobs:
...
           if [ ! -f "$src" ]; then
             echo "Expected build artifact '${src}' not found" >&2
             exit 1
           fi
 
           mkdir -p dist
-          mv "$src" "dist/${dest}"
+          cp "$src" "dist/sipnet"
+          tar -C dist -czf "dist/${archive_base}.tar.gz" sipnet
+          rm -f "dist/sipnet"
```
dlebauer, line 47:
```suggestion
          rm -f "dist/sipnet"
          echo "archive_base=${archive_base}" >> "$GITHUB_OUTPUT"
```

dlebauer (commented):
implementing suggestions from @Alomir

.github/workflows/release.yml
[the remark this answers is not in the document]
dlebauer: @copilot are we good here?

dlebauer:
@Copilot I do appreciate your enthusiasm, but when I ask "are we good" I mean is there anything left that I should do? Not "please implement a change without asking me". Can you please revert https://github.com/PecanProject/sipnet/commit/8c9ba57d84664439279cc7cad4e4365d30a87dd6

We don't need manual dispatch. We just want to upload when a release is published. If you really think any of the changes in that commit are worth keeping, add them as suggestions with explanations.

.github/workflows/release.yml
Alomir: Thanks!

Alomir (approved):
Looks good!

PR #320, opened 2026-04-20
[285] Rename --file-name to --file-prefix for clarity

### LINKED ISSUE #285: rename file-name to file-prefix for clarity
Mike noted in PR #276 that `input-file` takes a full filename while `file-name` is really a prefix. Follow up by renaming `file-name` to `file-prefix` for clarity, while keeping `file-name` as an alias.

Comment: https://github.com/PecanProject/sipnet/pull/276/changes#r2890991639

ANAMASGARD:
Fixes #285

## Summary
`--file-name` is misleading — SIPNET uses the value as a **prefix** for
multiple derived filenames (`.out`, `.NEE`, `.config`), not a single file.
This rename aligns the option name with its actual behavior.

## Changes
- Added `--file-prefix` as the primary long option (short: `-f`)
- Kept `--file-name` as a backward-compatible alias
- Updated help text and output placeholder strings in `cli.c`
- Updated `docs/user-guide/running-sipnet.md` and `model-inputs.md`
- Added changelog entry under `## [Unreleased]`

## Testing
- `make test` passes (unit + smoke)
- Manual parity: outputs from `--file-prefix`, `--file-name`, and `-f` are byte-identical


docs/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ sections to include in release notes:
 
 ### Changed
 
+- Renamed the CLI option `--file-name` to `--file-prefix` for clarity while keeping `--file-name` as a backward-compatible alias (#285)
```
Alomir: It might seem weird, but we use the PR number instead of the ticket number in the changelog (so this should be (#320)

src/common/context.c
```diff
@@ -180,10 +180,10 @@ void validateFilename(void) {
     printf("Error: fileName must be set for SIPNET to run\n");
     exit(EXIT_CODE_BAD_PARAMETER_VALUE);
   }
-  if (strlen(ctx.fileName) > FILENAME_MAXLEN - 10) {
+  if (strlen(ctx.fileName) > FILENAME_PREFIX_MAXLEN) {
```
Alomir: `ctx.fileName` should get renamed to `ctx.filePrefix`

docs/user-guide/model-inputs.md
```diff
@@ -212,7 +212,7 @@ Thus, command-line arguments override settings in the configuration file, and co
 | Option          | Default   | Description                                        |
 | --------------- | --------- | -------------------------------------------------- |
 | `input-file`    | sipnet.in | Name of input config file                          |
-| `file-name`     | sipnet    | Prefix of climate and parameter files              |
+| `file-name`     | sipnet    | Prefix of climate and parameter files (CLI: `--file-prefix`, config file: `file-name`) |
```
Alomir: This discrepancy is not good; we need to use (and document) `file-prefix` in the config file, but allow `file-name` for backwards compatibility.

Alomir (changes requested):
Thanks a lot for doing this! I had a few requests, let me know if anything is not clear.

Alomir (changes requested):
Changes look good, but must be incomplete. See the errors with the niwot and russell_2 smoke tests.

src/common/context.c
```diff
@@ -89,6 +87,10 @@ struct context_metadata *getContextMetadata(const char *name) {
   struct context_metadata *s;
   nameToKey(name);
   HASH_FIND_STR(ctx.metaMap, keyName, s);
+  // Backward compatibility for legacy FILE_NAME / file-name config keys.
+  if (s == NULL && strcmp(keyName, "filename") == 0) {
+    HASH_FIND_STR(ctx.metaMap, "fileprefix", s);
+  }
```
Alomir: You are going to need something similar in `updateCharContext` (which will fix the smoke test errors). 

One suggestion is to put this inside `nameToKey`, and have that function de-alias input key names. There might be a better place, though, look around.

docs/user-guide/running-sipnet.md
```diff
@@ -27,7 +27,8 @@ When the same option is specified in both places, **command-line arguments take
 | Option             | Short | Argument   | Default     | Description                                                                                |
 |--------------------|-------|------------|-------------|--------------------------------------------------------------------------------------------|
 | `--input-file`     | `-i`  | `<name>`   | `sipnet.in` | Name of input configuration file                                                           |
-| `--file-name`      | `-f`  | `<name>`   | `sipnet`    | Prefix for climate and parameter input files (looks for `<name>.clim` and `<name>.param`)  |
+| `--file-prefix`    | `-f`  | `<name>`   | `sipnet`    | Prefix for climate and parameter input files (looks for `<name>.clim` and `<name>.param`)  |
+| `--file-name`      | `—`   | `<name>`   | `sipnet`    | Backward-compatible alias for `--file-prefix`                                             |
```
Alomir: Let's move `--file-name` to a new table called Deprecated Options below ## Information Options

Alomir:
In `running-sipnet.md`, there are more changes needed:
- Input/Output Keys section
- Four example config files further down
Do a search for 'FILE_NAME' in that file.

Alomir (approved):
Looks good, thanks for your contribution!

PR #343, opened 2026-05-27
SIP310 Implement H2O mass-balance checking

### LINKED ISSUE #310: Implement H2O balance checking
PR #248 implemented carbon and nitrogen mass-balance checking.

This issue tracks implementing analogous H2O mass-balance checking.

For consistency with the current balance framework, this should compare the timestep change in tracked water stocks against modeled external water inputs and outputs.

Formulation:

delta(total tracked water) = external inputs - external outputs

where tracked water may include soilWater, snow, and any other explicit water pools (for example leaf water, if enabled), and external fluxes may include precipitation/rainfall, irrigation, transpiration, evaporation, immediate evaporation of intercepted water, sublimation, drainage, and fast flow as appropriate to the current implementation.

ANAMASGARD:
FIXES  #310
## Summary

Implements water mass-balance checking following the same framework as PR #248 (carbon/nitrogen):

- Tracks `soilWater` + `snow` each timestep (pre-update → post-update → post-clamp)
- Compares pool change against external water fluxes (rain, snowfall, irrigation, transpiration, evaporation, sublimation, drainage, fast flow)
- Logs a warning when `deltaWater` exceeds tolerance (`EPS`)
- Adds `bcdeltaC`, `bcdeltaN`, and `bcdeltaW` columns to `sipnet.out`
- Extends `testBalance` with `checkWater()` and updates smoke baselines

## Design notes

- **No leaf-water pool** — `ctx.leafWater` only caps `immedEvap`; excess interception goes to soil via `netRain`
- **`snowMelt` excluded** — internal transfer (snow → soil), not external I/O
- **`eventEvap` on both sides** — gross irrigation enters the system (input) and the evaporated portion leaves immediately (output); net soil addition is `eventSoilWater`

## Test plan

- [x] `make testbuild` — builds successfully
- [x] Water balance closes: `bcdeltaW = 0.00000` in smoke output; zero `"Water balance check failed"` warnings
- [x] `checkWater()` passes in all `testBalance` subtests
- [x] `make smoke` — passes after baseline update


Alomir:
> testBalance overall still fails due to a pre-existing carbon balance delta (deltaC ≈ 610 g C/m²), unrelated to this PR. Water checks pass.

@ANAMASGARD Can you say more about this? I don't see the failure.

ANAMASGARD:
@Alomir  sorry for the confusion, and also sorry my previous reply missed your actual question entirely (I answered the wrong thing).

You're right that you don't see the failure - because it wasn't a real pre-existing issue. That note was based on a testBalance failure I saw only in my local environment before syncing with master. After merging master, it disappeared completely and all tests pass cleanly in CI. I can't reproduce it anymore either.

It was likely a stale local build artifact rather than an actual code issue. I've removed that checkbox from the PR description to avoid further confusion.

PR #346, opened 2026-06-04
Update model inputs restrictions (tiny docs fix)

dlebauer:
Add restriction for 'carbon-saturation' requiring 'litter-pool'.

Docs only; too small to use PR template.


dlebauer (commented):
Looks great!

dlebauer:
bypassed review since it is small and going into open PR

PR #353, opened 2026-07-14
SIP350 Revamp C Buffer and Biomass N handling

Alomir:
### THIS PR HAS BEEN CLOSED

IT IS REPLACED BY #359.


tools/sipnet_view.py
```diff
@@ -140,52 +141,23 @@ def show_add_column_dialog(self) -> None:
...
-  parser.add_argument(
-    "-l",
-    "--layout",
-    choices=("subplots", "combined"),
-    default="subplots",
-    help="Initial plot layout. 'combined' uses twinned y-axes.",
+    help= (
+      "Path to a SIPNET output file. Defaults to ./sipnet.out. If directory, "
+      "looks for sipnet.out there.",
+    )
```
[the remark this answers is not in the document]
Alomir: 👍

src/sipnet/debug_log.c
```diff
@@ -55,11 +58,19 @@ void initDebugArrays() {
...
+  debugFields->enviDF[ind++] = (DebugField){"plantWoodN", DEBUG_FIELD_DOUBLE, &envi.plantWoodN},
+  debugFields->enviDF[ind++] = (DebugField){"plantLeafN", DEBUG_FIELD_DOUBLE, &envi.plantLeafN},
+  debugFields->enviDF[ind++] = (DebugField){"coarseRootN", DEBUG_FIELD_DOUBLE, &envi.coarseRootN},
+  debugFields->enviDF[ind++] = (DebugField){"fineRootN", DEBUG_FIELD_DOUBLE, &envi.fineRootN},
   debugFields->enviDF[ind++] = (DebugField){"minN", DEBUG_FIELD_DOUBLE, &envi.minN},
   debugFields->enviDF[ind++] = (DebugField){"soilOrgN", DEBUG_FIELD_DOUBLE, &envi.soilOrgN},
   debugFields->enviDF[ind++] = (DebugField){"litterN", DEBUG_FIELD_DOUBLE, &envi.litterN},
   debugFields->enviDF[ind++] = (DebugField){"plantStorageN", DEBUG_FIELD_DOUBLE, &envi.plantStorageN},
-  debugFields->enviDF[ind  ] = (DebugField){"plantWoodCStorageDelta", DEBUG_FIELD_DOUBLE,&envi.plantWoodCStorageDelta};
+  debugFields->enviDF[ind++] = (DebugField){"plantWoodCStorageDelta", DEBUG_FIELD_DOUBLE,&envi.plantCAccountingDelta};
```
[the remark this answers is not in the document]
Alomir: 👍

Alomir:
@copilot please open a PR to update the unit tests in tests/sipnet to cover these changes. Create new similar tests if needed.

src/sipnet/sipnet.c
```diff
@@ -1698,6 +1697,45 @@ void updatePoolsForSoil(void) {
...
+    logInfo("Plant mortality detected: wood or coarse root carbon is zero or "
+            "negative (woodC %f, coarseRootC %f year %d day %d time %6.3f);"
+            "zeroing out biomass pools\n",
+            envi.plantWoodC, envi.coarseRootC, climate->year, climate->day,
+            climate->time);
+    // TODO: Add check for harvest cause, update info to say so; convert to
+    //       warning if not
+
+    // Force pools to zero
+    // TODO: add positive remnants to litter/soil as appropriate?
```
dlebauer: Yes, must conserve mass!

src/sipnet/sipnet.c
```diff
@@ -1569,7 +1555,23 @@ void initPhenologyTrackers(void) {
                                                // this year
 }
 
+void initPlantSurvivalTracker(void) {
+  plantSurvival.justDied = 0;
+  if (envi.plantWoodC > TINY && envi.coarseRootC > TINY) {
```
dlebauer: I think that it is or should be valid to have zero allocation to coarse roots.

How about `(envi.plantWoodC + envi.coarseRootC > TINY)`?

Alomir: @dlebauer How about `(fineRoot + coarseRoot) >= 0`? No roots should mean plant death, right?

src/common/util.h
```diff
@@ -13,6 +13,10 @@
 
 #define TINY 0.000001  // to avoid those nasty divide-by-zero errors
 
+// Floating-point precision threshold for balance checks and clamping warnings.
+// Differences smaller than this are treated as numerical noise.
+#define EPS 1e-8
```
infotroph, line 18: Would it make sense to combine EPS and TINY into one value?

Alomir: @infotroph I view them as serving different purposes, and think that perhaps they should move farther apart...

I think of TINY as "anything below this amount is practically zero, let's treat it as such and not propagate such small quantities around". Perhaps this should be bigger than 1e-6, maybe e-5 or e-4.

EPS is "numerical roundoff has given us a non-zero quantity that should actually be zero". This could probably be much smaller than 1e-8, say e-12 or even smaller.

Alomir: One could also infer that I think the comment on TINY above is incorrect - or, at least, only one aspect. However, that comment is making me think that 1e-6 is good.

src/sipnet/sipnet.c
dlebauer: Do you mean ">" instead of ">="? Else If = 0, then die.

src/sipnet/sipnet.c
```diff
@@ -1698,6 +1703,82 @@ void updatePoolsForSoil(void) {
...
+    envi.plantWoodC = 0.0;
+    envi.plantLeafC = 0.0;
+    envi.coarseRootC = 0.0;
+    envi.fineRootC = 0.0;
+    envi.plantCAccountingDelta = 0.0;
+    if (ctx.nitrogenCycle) {
+      envi.plantWoodN = 0.0;
+      envi.plantLeafN = 0.0;
+      envi.coarseRootN = 0.0;
+      envi.fineRootN = 0.0;
```
dlebauer, line 1772: current: looks like an oversight. plantStorageN is transferred to litterN above but not zeroed here with the other plant N pools, duplicating that N.

proposed: zero it out here

```suggestion
      envi.fineRootN = 0.0;
      envi.plantStorageN = 0.0;
```

src/sipnet/limitations.c
dlebauer, line 97: I think this will be obsolete after the proposed simplification to a single, non-biomass `extraN` storage pool, but I wanted to note here: with organ-specific `extraN`, plant N demand is non-linear in the creation fluxes. Thus, reducing the C creation flux and then multiplying the original fixation and uptake fluxes by the same reduction factor will not produce the realized N demand.

A minimal interim fix may be to scale the creation fluxes and then call calcNFixationAndUptakeFluxes() again, so fixation and uptake are recalculated from the reduced growth.

```suggestion
fluxes.woodCreation *= reduction;
fluxes.leafCreation *= reduction;
fluxes.fineRootCreation *= reduction;
fluxes.coarseRootCreation *= reduction;

calcNFixationAndUptakeFluxes();
```


This is a nice example of where the simplification should make the model structure easier to reason about and reduce the surface of possible bugs.

dlebauer (commented):
Reviewing with the goal of getting this merged as an interim approximation of the proposed single-pool simplification, and identified two changes required for correctness, with the help of ChatGPT.

I left two small comments and proposed associated fixes. Please reject if it is more complicated than I imagine - the refactor is the priority.

src/sipnet/sipnet.c
Alomir: Good catch! This is new code that is still untested, as that most certainly would fail the mass balance check.

Alomir:
CLOSING THIS IN FAVOR OF #359

PR #359, opened 2026-08-01
SIP350 Update carbon accounting + plant mortality

### LINKED ISSUE #325: Insufficient carbon for growth should cause mortality, and fix C balance
## Problem

SIPNET can generate negative values of biomass. When biomass <0, plants should die.

In the current code, impossible negative states biomass pools are repaired after they are updated by clamping stocks to zero rather than handling plant carbon-deficit failure explicitly. 

However, when plantWoodCStorageDelta term added to the wood update, it is possible for aboveground biomass to be negative at the end of the time step (and thus, in the sipnet.out).

Related but out of scope:
If the clamping is > tolerance, the [`ensureNonNegative`](https://github.com/divine7022/sipnet/blob/1bd16b782c9941c98abcb9615e8626f1fd78c309/src/sipnet/sipnet.c#L1631-L1640) function warns but still clamps to 0. It seems this could introduce phantom mass.

## Proposed change

Implement a mortality rule:
- do not allow live structural biomass pools (wood, leaf, fine root, root) to go negative (current clamping behavior)
- do not let negative wood storage make wood-based fluxes or biomass outputs negative
- trigger mortality before stock clamping if a timestep would drive any live structural biomass pool negative
- on mortality, set `plantWoodC`, `plantLeafC`, `fineRootC`, `coarseRootC`, and `plantWoodCStorageDelta` to zero in the same timestep
- transfer the biomass of the dead plants as currently implemented:
    - roots --> soil
    - leaves and wood --> litter if litter-pool=on, otherwise --> soil
- if `plantWoodCStorageDelta` is negative at death, remove that from the amount transferred to litter / soil 
- do not rely on `clampedC` / `clampedN` in the mortality path

## Acceptance criteria

- No live structural biomass pool < -eps [eps = small tolerance].
- Carbon-deficit triggers explicit mortality instead of clamp-based repair.
- Mortality conserves mass and uses current aboveground/belowground routing conventions.
- If wood storage is negative at death, mortality routing does not create phantom carbon. I can't think of anything simpler than mass balance is maintained by taking from other pools either evenly or in order: wood, root, [leaf + fine root])

## Alternatives

After looking into this, I realized how complex it is to represent mortality! In croplands especially, mortality is not a high priority. If it is simpler to throw an error, lets tag this issue 'on hold' and just do that. 

## Related

- #324 may be resolved by the fix here
- #282
- #248

### LINKED ISSUE #350: Revamp biomass N handling
### Carbon Accounting
- When mean NPP is negative, we now distribute that negative value (loss) over the biomass pools according to the standard NPP allocation percents
  - As a result, the C accounting buffer no longer descends into large negative territory, and stays within zero as it should
  - If this causes leaf C to become negative, we deduct that quantity from wood C
  - Similarly, fine and coarse roots check for negative values, and shift accordingly
- The nitrogen associated with this carbon loss is added to the nitrogen storage pool (`envi.plantStorageN`) via a new flux
  - Carbon loss from "negative growth" (mean NPP < 0) only removes carbon; the nitrogen left behind is available to be used by growth (including leaf-on events)

### Plant Mortality
We now check for plant death and reemergence

- Plant death is determined if either wood C or combined root C become zero or less
- Plant reemergence is determined when the above is no longer true
- On death, all remaining biomass (both C and N) are moved to the appropriate litter/soil pools

### Other

- Replace `plantWoodCStorageDelta` with hopefully-clearer `plantCAccountingDelta`
- Update and add unit test coverage of changes
- Update debug logging
- Refactor `sipnet-view` and `sipnet-debug-view`

Alomir:
## Summary

- **What**: Revamp handling of the C pseudo-storage pool
- **Motivation**: Current handling is broken.

NOTE: THIS REPLACES #353 

- Existing unit tests updated
- New unit tests added to cover new functionality
- Smoke tests run and updated; diff results below

[smoke_check.txt](https://github.com/user-attachments/files/30669876/smoke_check.txt)

## Related issues

- Fixes #350, #325

## Checklist

- [x] Related issues are listed above. [PRs without an approved, related issue may not get reviewed](docs/CONTRIBUTING.md#propose-and-receive-feedback).
- [x] PR title has the issue number in it ("[#<number>] \<concise description of proposed change>")
- [x] Tests added/updated for new features (if applicable)
- [x] `docs/CHANGELOG.md` updated with noteworthy changes
- [x] Code formatted with `clang-format` (run `git clang-format` if needed)

---


Alomir:
@copilot please open a PR to update the unit tests in tests/sipnet to cover these changes. Create new similar tests if needed.

tools/sipnet_view_common.py
```diff
@@ -499,6 +507,96 @@ def format_datetime_for_display(value: datetime) -> str:
...
+      near_idx = find_closest_sorted(xd, ed)
+      #print(xd[near_idx], yd[near_idx])
+      x = np.datetime_as_string(xd[near_idx], unit='m')
+      y = yd[near_idx]
+      #breakpoint()
+      # Set a custom message on the toolbar status bar
+      self.set_message(f"Closest data point:\nX: {x}  Y: {y}")
+    else:
+      # Fallback for when the mouse leaves the plot area
+      self.set_message("Outside plotting area")
```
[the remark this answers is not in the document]
Alomir: Fixes manually

src/sipnet/restart.c
```diff
@@ -132,107 +141,163 @@ typedef struct RestartState_s {
...
+    logInternalError("Restart array size mismatch: metaPF\n");
+    exit(EXIT_CODE_INTERNAL_ERROR);
+  }
+
+  ind = 0;
+  state->schemaPF[ind++] = (StateField){"schema_layout.envi_size",               FT_SPECIAL, 0, RESTART_SCHEMA_LAYOUT_ENVI_SIZE};
+  state->schemaPF[ind++] = (StateField){"schema_layout.trackers_size",           FT_SPECIAL, 0, RESTART_SCHEMA_LAYOUT_TRACKERS_SIZE};
+  state->schemaPF[ind++] = (StateField){"schema_layout.phenology_trackers_size", FT_SPECIAL, 0, RESTART_SCHEMA_LAYOUT_PHENOLOGY_TRACKERS_SIZE};
+  state->schemaPF[ind++] = (StateField){"schema_layout.event_trackers_size",     FT_SPECIAL, 0, RESTART_SCHEMA_LAYOUT_EVENT_TRACKERS_SIZE};
+  state->schemaPF[ind++] = (StateField){"schema_layout.invalid",                 FT_INVALID, NULL, FIELD_INVALID};
```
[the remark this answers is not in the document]
Alomir: Fixed manually

tests/sipnet/test_sipnet_infrastructure/testDebugLogFiles.c
```diff
@@ -136,12 +128,55 @@ int run(void) {
...
+  int status = 0;
 
   logTest("Starting testDebugLogFiles\n");
 
-  status = run();
+  status |= init();
+
+  status |= run();
+
+  status |= cleanup();
```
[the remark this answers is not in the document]
Alomir: Somewhat disagree; changed manually

tools/sipnet_debug_view.py
```diff
@@ -192,66 +195,26 @@ def show_add_column_dialog(self) -> None:
...
-    type=int,
-    default=6,
+    "--title",
+    default="Sipnet Debug Log Viewer",
     help=(
-      "Reserved for future use. Accepted for compatibility, "
-      "but warnings are currently disabled."
+      "Title displayed on the viewer window."
     ),
   )
```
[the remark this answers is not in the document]
Alomir: Good catch! Fixed manually.

src/sipnet/events.c
```diff
@@ -166,6 +166,11 @@ EventNode *createEventNode(int year, int day, int eventType,
       }
       newEvent->eventParams = lParams;
     } break;
+    case PLANTDEATH: {
+      logWarning("PLANTDEATH event found for year %d day %d, but not "
+                 "implemented as an input event; ignoring\n",
+                 gEvent->year, gEvent->day);
+    } break;
```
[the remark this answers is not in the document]
Alomir: Changed to error.

src/sipnet/restart.c
```diff
@@ -34,7 +36,8 @@
 #define NUM_ENVI_FIELDS 13
 #define NUM_TRACKER_FIELDS 32
 #define NUM_PHENOLOGY_TRACKERS_FIELDS 3
-#define NUM_EVENT_TRACKERS_FIELDS 1
+#define NUM_SURVIVAL_TRACKERS_FIELDS 1
+#define NUM_EVENT_TRACKERS_FIELDS 3
```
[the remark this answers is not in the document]
Alomir: Done in change below.

Alomir:
Close/reopen to restart checks

src/sipnet/events.c
```diff
@@ -166,6 +166,11 @@ EventNode *createEventNode(int year, int day, int eventType,
       }
       newEvent->eventParams = lParams;
     } break;
+    case PLANTDEATH: {
```
dlebauer, line 169: other `logErrors` visible here are followed by `exit()`, should this be the case here as well?

src/sipnet/balance.c
dlebauer, line 97: @copilot please check if math is correct here. confirm whether mass is correctly conserved, or if it is created or destroyed due to an accounting error.

src/sipnet/sipnet.c
```diff
@@ -1698,6 +1714,94 @@ void updatePoolsForSoil(void) {
...
 }
 
+/*!
+ * Simple check for plant death, to perform needed cleanup.
+ *
+ * This avoids mistaken warnings in succeeding steps. Also checks for plant
+ * reemergence (say, after a planting)
+ */
+void checkForMortality(void) {
+  double totalWoodC = getTotalWoodC();
```
dlebauer: @copilot Could checking total wood/root C here miss an individual pool that has gone negative but is offset by another pool or by plantCAccountingDelta?

src/sipnet/events.c
```diff
@@ -584,12 +601,12 @@ void processEvents(void) {
         fluxes.eventOutputC += outputC / climLen;
         if (ctx.nitrogenCycle) {
           // just plantWoodC here, not woodC
-          outputN = ((envi.plantWoodC / params.woodCN +
-                      envi.plantLeafC / params.leafCN) *
-                         fracRA +
-                     (envi.fineRootC / params.fineRootCN +
-                      envi.coarseRootC / params.woodCN) *
-                         fracRB);
+          outputN = (envi.plantWoodC / params.woodCN +
```
dlebauer, line 611: should this calculation account for the accounting delta, like https://github.com/PecanProject/sipnet/pull/359/changes#diff-bd222092a67702c305456d7fdcc76171ed32f16532fe8912ef51d5b019d72af6R554?

docs/CHANGELOG.md
```diff
@@ -31,15 +31,19 @@ sections to include in release notes:
...
+- New `sipnet-debug-view` tool for visualizing SIPNET debug output files (#359)
+- Plant mortality check with event output on occurrence (#359)
 
 ### Fixed
 
 ### Changed
 
 - Renamed the CLI option `--file-name` to `--file-prefix` for clarity while keeping `--file-name` as a backward-compatible alias (#320)
 - Values in `events.out` changed to pool deltas rather than flux amounts (#349)
+- Renamed `nppStorage` output column to `nppDelta` to clarify that it is a delta rather than a pool value (#359)
```
dlebauer: smoke test still has `nppStorage`

tools/sipnet_view_common.py
```diff
@@ -499,6 +507,103 @@ def format_datetime_for_display(value: datetime) -> str:
...
+      if not lines:
+        super().mouse_move(event)
+        return
+      line = lines[0]
+      xd = line.get_xdata()
+      yd = line.get_ydata()
+      target_dt = mdates.num2date(event.xdata).replace(tzinfo=None)
+      target = np.datetime64(target_dt)
+      near_idx = find_closest_sorted(xd, target)
+      x = pd.to_datetime(xd[near_idx]).strftime('%Y-%j %H:%m')
```
dlebauer: **M**inutes instead of **m**onths

```suggestion
      x = pd.to_datetime(xd[near_idx]).strftime('%Y-%j %H:%M')
```

src/sipnet/events.c
```diff
@@ -534,7 +548,10 @@ void processEvents(void) {
         const double fracRB = harvParams->fractionRemovedBelow;
         const double fracTB = harvParams->fractionTransferredBelow;
 
-        const double woodC = envi.plantWoodC + envi.plantWoodCStorageDelta;
+        eventTrackers.harvestFracRemoved = fracRA + fracRB;
+        eventTrackers.harvestFracTransferred = fracTA + fracTB;
```
dlebauer: as written, `eventsTrackers.harvestFracRemoved` and `harvestFracTransferred` can range from 0-2. Would it be clearer to track the aboveground and belowground fractions separately? Okay by me to merge and file a new issue.

dlebauer (changes requested):
Excited to have this so close to merging!

I've put a few comments. Please defer any that don't affect values in sipnet.out or restart file (e.g. docs changes, fracHarvTransfer in events.in) directly so that we can work with the correct version and track these fixes/improvements separately

Required docs changes include:
- `survival.isAlive` to `docs/developer-guide/restart-checkpoint.md:31` 
- new equations
- other?

FYI I tagged copilot in some comments and also used chatgpt to help track down and understand these. I didn't realize that tagging copilot would have it look at all comments, but 🤞🏻 the 🤖 is helpful! update: I didn't realize it would be too helpful - it applied changes without review. So I asked it to revert the patches and submit as a new PR, which apparently it hasn't done. If it comes to it revert to where I started reviewing - https://github.com/PecanProject/sipnet/pull/359/commits/8b2fe9924a9c1e2be2e263f75ca26752613f289c

src/sipnet/balance.c
dlebauer: @copilot The balance closes mathematically, but does clamping a negative pool to zero still create mass? Adding it to inputsN only labels that mass as an external input. Is clamping guaranteed to be limited to rounding error?

src/sipnet/sipnet.c
dlebauer: Thanks for the confirmation: ensureNonNegativeStocks() zeros the negative sub-pool without adjusting the flux, so the clamp can add mass. It seems that per-pool fluxes should be bounded before the update. please confirm @Alomir

src/sipnet/nitrogen.c
```diff
@@ -175,7 +185,9 @@ void updateNitrogenPools(void) {
   double uptake = fluxes.nUptake * climate->length;
   double uptakeFromStorage = fmin(uptake, envi.plantStorageN);
   envi.plantStorageN +=
-      fluxes.leafOffNResorption * climate->length - uptakeFromStorage;
+      (fluxes.leafOffNResorption + fluxes.reductionNResorption) *
+          climate->length -
+      uptakeFromStorage;
```
dlebauer: **resolve intended behavior and documentation of n uptake fluxes**

Blockers:
- [ ] confirm intended behavior:
  - (current) storage used after calculating n fixation or
  - (alternative) storage satisfies demand before calculating n fixation
- [ ] fix if appropriate

Defer:
- [ ] align definitions of nUptake, uptakeFromStorage, uptakeFromMinN
  - replace nUptake w/ uptakeFromMinN: actual soil uptake
  - also define (at least in .md):
    - nUptake / uptake: total non-fixation N demand being satisfied
    - uptakeFromStorage: internal plant N transfer
    - uptakeFromMinN

Getting behavior correct is priority - semantics and docs are non-blocking; can be moved to separate issue. 

#### Docs define nUptake as mineral N taken from soil: 

**these definitions apply to `uptakeFromMinN`**

- https://github.com/PecanProject/sipnet/blob/cf3987580d27a57c70884cf55fe1139155c06a67/src/sipnet/state.h#L579
- https://github.com/PecanProject/sipnet/blob/cf3987580d27a57c70884cf55fe1139155c06a67/src/sipnet/state.h#L721
- https://github.com/PecanProject/sipnet/blob/cf3987580d27a57c70884cf55fe1139155c06a67/docs/user-guide/model-outputs.md#L45

#### implementation - nUptake is from minN and storageN

- https://github.com/PecanProject/sipnet/blob/cf3987580d27a57c70884cf55fe1139155c06a67/src/sipnet/nitrogen.c#L186
- https://github.com/PecanProject/sipnet/blob/cf3987580d27a57c70884cf55fe1139155c06a67/src/sipnet/nitrogen.c#L195

src/sipnet/events.c
dlebauer: the accounting delta should not itself carry N. My concern is whether calculating woodDelta from total wood and later applying it entirely to plantWoodC keeps C and N removal consistent when the accounting delta is nonzero.

src/sipnet/balance.c
Alomir: @dlebauer - clamping (non-roundoff) values absolutely creates mass. This is a known issue (though not tracked! which is likely the real problem) waiting for us to decide what to do about it.

Eventually, I'm hoping we make this an error, and we can remove the clamping logic from the balance checks. It was put in there as... a stop-gap, I guess? while we hardened the system, to allow SIPNET to run even when these issues arose.

src/sipnet/sipnet.c
Alomir: @dlebauer  - good questions!

I think this is a spec question. What behavior are we looking for here for wood C? (Note: I'm pretty sure the other pools are handled - or it's a bug if ensureNonNegative gets triggered for them). 

Or, is wood C not really the issue here, and we want a general "check and deal with fluxes that cause pools to go negative" function? (Which would be a very new thing - not against it at all, sounds great, but SIPNET hasn't really done that at all beyond ensureNonNegative.)

src/sipnet/events.c
Alomir: Yes indeed; originally a warning, and I forgot the exit when I changed it to an error

src/sipnet/events.c
Alomir: Yep, copilot nailed that one

src/sipnet/events.c
Alomir: Oooh, right, this is more complicated. The point here was to indicate how much of the total biomass was removed and transferred, which I think is calculable. Splitting into above and below separately I _think_ would result in just repeating the harvest event input params.

src/sipnet/sipnet.c
Alomir: One conclusion:
* plantWoodC <= 0 means death
* totalWoodC <= 0 means death
Check both.

src/sipnet/events.c
Alomir: Will fix.

src/sipnet/nitrogen.c
Alomir: Switch to "(alternative) storage satisfies demand before calculating n fixation"

Alomir: Make naming more consistent with (new) implementation

src/sipnet/balance.c
Alomir: Created #365 to track this eventual task

src/sipnet/sipnet.c
Alomir: Done.

docs/CHANGELOG.md
Alomir: Apparently, this change got reverted - it's still called nppStorage. I'm going to leave questions of renaming this output to a follow-up task.

tools/sipnet_view_common.py
Alomir: Fixed

src/sipnet/events.c
Alomir: Done.

Alomir:
Asking for copilot code review here, but I expect it to complain about unit tests. We'll see!

src/sipnet/events.c
Alomir: Fixed

src/sipnet/nitrogen.c
Alomir: Hopefully fixed!

src/sipnet/sipnet.c
```diff
@@ -1698,6 +1721,93 @@ void updatePoolsForSoil(void) {
...
+          "year %d day %d time %6.3f; zeroing out biomass pools\n",
+          envi.plantWoodC, totalWoodC, envi.coarseRootC, envi.fineRootC,
+          climate->year, climate->day, climate->time);
+    }
+
+    // Move remnants to litter/soil as appropriate. Note we add all pools, even
+    // if some are negative, as we should for mass balance. Given that there are
+    // multiple processes being modeled, it's believable that there may be a bit
+    // of overshoot when a plant dies - for example, a 100% harvest event with
+    // any overall loss (respiration, turnover).
```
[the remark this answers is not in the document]
Alomir: I think I disagree? We are conserving mass, it's ensureNonNegative() that's foiling it. Ticket #365 should address this.

src/sipnet/nitrogen.c
```diff
@@ -145,11 +146,32 @@ double calcNFixationFrac(void) {
...
+  double storageFlux =
+      fmax(0.0, envi.plantStorageN / climate->length - leafOnNFlux);
+
+  // Remaining demand for uptake/fixation
+  double remDemandFlux = fmax(0.0, nDemandFlux - storageFlux);
+  // Now parcel that out between fixation and uptake
   double nFixationFrac = calcNFixationFrac();
+  fluxes.nFixation = nFixationFrac * remDemandFlux;
+  fluxes.nUptake = (1 - nFixationFrac) * remDemandFlux;
+}
```
dlebauer, line 168: It appears that permitted growth is based on current-step resorption, but soil uptake is not. The resorbed N can then remain in plantStorageN while the same N is drawn from minN. Should current resorption be included in storageFlux?

dlebauer (changes requested):
🥇 
One inline question about N calculation logic. I will share some remaining thoughts in slack that should not hold up this PR.

src/sipnet/nitrogen.c
Alomir: @dlebauer "permitted growth" as calculated where?

I actually had it that way at first... but I spent a bit of time here thinking about the "core" of how SIPNET handles things: calculate fluxes based on current state, then update pools. Every time we calculate a flux *based on a current-step flux*, I feel that we should consider if we might be violating that. There are places where it's absolutely the correct thing - e.g., using C fluxes to calculate N fluxes is just inherent in our N model (though actually, we could probably do those from actual state, it would just be longer and result in the same outcome).

Here, specifically, I think we are skirting the edge of the SIPNET methodology no matter what, as this feels like a big "look ahead" to what the state is going to be at the end of the timestep. However, our desire to consider the storage N before soil N to satisfy demand necessitates some level of doing this. And, in that vein, bringing in all of the demand terms makes sense to me (and keeps the storage from going negative). However, considering what's going to come INTO the pool in this time step feels akin to, say, taking into account this step's leaf growth when calculating leaf litter (which of course we don't do).

Alomir: If by "permitted growth" you mean our limitation checks, then I feel like those were an exception that we (or at least I, hah hah) grudgingly accepted; I'd rather change those, than let the model drift further from the "core" I mentioned above.

But, if you mean something else, let me know! That would be something to discuss.

dlebauer: Thanks for clarifying, it makes sense to change the limitation checks, but I am still concerned about cases that allow minN < 0. I'll follow up in Slack.

Alomir: To summarize:
* Using incoming plantStorageN (from the resorption terms) as plantAvailableN is not working, since it is not available to offset minN losses directly in the same time step
  * At least, not without major shenanigans that likely violate the calcAlllFluxes->updateAllPools contract
  * Not addressing this leads to states where end the step with `minN = -x` and `plantStorageN = +x`, where `x` is that incoming plantStorageN
* We can still "look ahead" for the minN pool directly in calcNitrogenLimitation, but we need to stop doing the same for plantStorageN

Alomir: Also also, @dlebauer + chatGPT/claude did an excellent job of _finding_ this issue

src/sipnet/nitrogen.c
```diff
@@ -109,13 +110,11 @@ double calcPlantAvailableN(void) {
...
   double leafOnCFlux = fluxes.leafOnCreation + fluxes.eventLeafOnCreation;
   double leafOnNFlux = calcLeafOnNFromC(leafOnCFlux);
-  // Note: leafOnNFlux has already been limited to not exceed plantStorageN
-  double leafOffNFlux =
-      fluxes.leafOffNResorption + fluxes.eventLeafOffNResorption;
-  double unclaimedStorage =
-      envi.plantStorageN + (leafOffNFlux - leafOnNFlux) * climate->length;
+  double unclaimedStorage = envi.plantStorageN - leafOnNFlux * climate->length;
   double nonUptakeDelta = calcMinNNonUptakeFluxes() * climate->length;
   return fmax(0.0, envi.minN + unclaimedStorage + nonUptakeDelta);
```
dlebauer, line 119: @copilot Please analyze, don’t modify. Is combining storage and mineral N here consistent with storage-first uptake in checkNitrogenLimitation()? Please test partial storage with nonzero fixation and check whether updateNitrogenPools() can leave minN < 0.

Alomir:
@copilot please repeat this analysis, again without modifying. Is combining storage and mineral N here consistent with storage-first uptake in checkNitrogenLimitation()? Please test partial storage with nonzero fixation and check whether updateNitrogenPools() can leave minN < 0

dlebauer (approved):
🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳
🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳
🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳
🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳
🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳
🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳

